Issue #11, opened 2021-11-02
Implement verification of `dev.sigstore.cosign/certificate`


flavio:
## Description

The verification code should be extended to verify the contents of the `dev.sigstore.cosign/certificate` annotation.

This annotation is produced by `cosign` when signing in keyless mode:

```console
COSIGN_EXPERIMENTAL=1 cosign sign registry-testing.svc.lan/kubewarden/disallow-service-nodeport
```

This will produce a manifest like that:

```json
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "size": 233,
    "digest": "sha256:58a221cb6acea000c4a38491f6f1a2533c095a39a37f0af9275554a46fe69653"
  },
  "layers": [
    {
      "mediaType": "application/vnd.dev.cosign.simplesigning.v1+json",
      "size": 277,
      "digest": "sha256:e7801de935512e22b237c7b57e42ca4d20e2134bdf1b98f976c8f55d9cfe4067",
      "annotations": {
        "dev.cosignproject.cosign/signature": "MEUCIGqWScz7s9aP2sGXNFKeqivw3B6kPRs56AITIHnvd5igAiEA1kzbaV2Y5yPE81EN92NUFOl31LLJSvwsjFQ07m2XqaA=",
[...]

Issue #9, opened 2021-11-02
Provide way to download rekor public key from Sigstore TUF repository


flavio:
Currently library end users have to provide the Rekor public key when creating a Cosign Client. This is required to verify the cosign bundles.

Having to specify the Rekor public key is needed when the default public Rekor instance is not used (e.g. for testing, air-gapped environments,...), however this is not what the majority of people will do.

It would be of great help to provide a some method that downloads the official Rekor public key from Sigstore's TUF repository. This is what `cosign` does behind the scenes.

flavio:
We need a Rust library that implements TUF protocol in order to download the public key from Sigstore's TUF repository.

I think [tough](https://github.com/awslabs/tough) is a good candidate because it's considered stable and is well maintained.

However, I found some issues while trying to use it, I sent some patches upstream:

  * https://github.com/awslabs/tough/pull/425
  * https://github.com/awslabs/tough/pull/426

flavio:
Currently blocked because of external issues on the tough crate

flavio:
Quick update: the issue inside of tough seems to be slowly moving forward. In the meantime I have a branch which is implementing the retrieval of the keys from the TUF repository using the patched tough library.

I'll share more progresses later

PR #12, opened 2021-11-13
Implement keyless verification

flavio:
#### Summary

Implement verification of objects signed in keyless mode.

#### Ticket Link

Fixes https://github.com/sigstore/sigstore-rs/issues/11


src/cosign/signature_layers.rs
```diff
@@ -41,210 +38,254 @@ use crate::{crypto::CosignVerificationKey, simple_signing::SimpleSigning};
...
+            None => None,
+        };
+        Ok(certificate_key)
+    }
+
+    /// Checks whether the SignatureLayer's can be verified using the given `verification_key`.
+    ///
+    /// When no `verification_key` is given, the verification will be done using the Certificate
+    /// and the Bundle that are eventually part of the SignatureLayer. If none of them is available,
+    /// the verification will fail.
```
lkatalin, line 175: Is this why the verification key is now an optional parameter?

lkatalin (commented):
I think this LGTM. Thanks for implementing this feature! I especially appreciate the thorough testing and documentation. Would it make sense in the documentation to point people to how they can retrieve the values for the Rekor pubkey (for the public instance) and Fulcio cert? I realized I am not sure of that.

src/cosign/signature_layers.rs
flavio: Yes, that's correct.

flavio:
> Would it make sense in the documentation to point people to how they can retrieve the values for the Rekor pubkey (for the public instance) and Fulcio cert? I realized I am not sure of that.

My personal goal is to provide an API inside of this library to handle that. Unfortunately this is currently blocked by the Sigstore TUF repository containing "broken" (as in, not following the strict rules of Rust) data.
This is tracked here: https://github.com/sigstore/sigstore-rs/issues/9

The good news is that @asraa just got a fix merged (see https://github.com/sigstore/root-signing/pull/52). I think I can provide this API and then we can tag a new release of sigstore-rs which will have both this code, and the one that can fetch rekor+fulcio data via TUF

lkatalin:
> > Would it make sense in the documentation to point people to how they can retrieve the values for the Rekor pubkey (for the public instance) and Fulcio cert? I realized I am not sure of that.
> 
> My personal goal is to provide an API inside of this library to handle that. Unfortunately this is currently blocked by the Sigstore TUF repository containing "broken" (as in, not following the strict rules of Rust) data. This is tracked here: #9
> 
> The good news is that @asraa just got a fix merged (see [sigstore/root-signing#52](https://github.com/sigstore/root-signing/pull/52)). I think I can provide this API and then we can tag a new release of sigstore-rs which will have both this code, and the one that can fetch rekor+fulcio data via TUF

Ah, I see! That's a great plan. Thanks @flavio

lkatalin (approved):

PR #13, opened 2021-12-02
Add option to build with rustls-tls support

ereslibre:
#### Summary

Allow to choose the SSL backend. By default it's `native-tls`, but it can be set to `rustls-tls` in order to not depend on the OpenSSL stack of the system.

```release-note
Allow to choose the SSL stack by adding a `native-tls` or `rustls-tls` feature
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,13 +7,17 @@ authors = [
   "sigstore-rs developers",
 ]
 
+[features]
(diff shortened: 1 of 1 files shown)


flavio (approved):
LGTM

flavio:
The clippy error is unrelated with your changes. I'll fix that later on

PR #14, opened 2021-12-03
Make accept_invalid_hostnames conditional

ereslibre:
Depends on: https://github.com/sigstore/sigstore-rs/pull/13
Depends on a release of the `oci-distribution` crate that contains https://github.com/krustlet/oci-distribution/pull/8.

Make `accept_invalid_hostnames` optional, given it's only used in the `openssl` backend. When `oci-distribution` is released with the mentioned commit (ca1fc724308ed183b42b3239e49c1b23ef183524), the `accept_invalid_hostnames` will **only and only** exist in the struct if the chosen SSL backend is `openssl`.

Signed-off-by: Rafael Fernández López <ereslibre@ereslibre.es>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ anyhow = "1.0.44"
 async-trait = "0.1.51"
 base64 = "0.13.0"
 ecdsa = { version = "0.12.4", features = ["verify", "pem", "der", "pkcs8"] }
-oci-distribution = { version = "0.7.0", default-features = false }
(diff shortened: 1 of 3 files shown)


flavio (approved):
LGTM, let's wait for a new release of the oci-distribution crate to be released

flavio:
@ereslibre Can you update this PR to use oci-distribution 0.8.1? This is a freshly release that includes the fix you mentioned

ereslibre:
Updated.

Issue #17, opened 2021-12-20
verify example is confusing


lukehinds:
https://github.com/sigstore/sigstore-rs/tree/main/examples/verify

I don''t understand why someone would provide a a cert email, fuclio cert and rekor pub key to verify the signing made with a cosign generated key. There is no trust relation between the cosign.pub key and the others.

I would have thought this would be more relevant:

`cargo run --example verify -- -k cosign.pub  registry-testing.svc.lan/busybox`

or for keyless:

```
cargo run --example verify -- \
  --rekor-pub-key ~/.sigstore/root/targets/rekor.pub \
  --fulcio-crt fulcio.crt.pem \
  registry-testing.svc.lan/busybox
```

flavio:
Right now the library is extremely conservative. When looking at all the signature layers of a cosign signature, the library attempts to verify the trustworthiness of each layer:

  * Layers produced without rekor integration enabled and with "traditional" key signing are always trusted
  * Layers produced with rekor integration and "traditional" key signing are trusted only if the rekor bundle is verified. For that we need the public key of rekor
  * Layers produced with keyless signing are trusted only if the rekor bundle is verified and if the certificate has been issued by fulcio's CA. For that we need both the public key of rekor and fulcio's certificate.

The trusted layers are then further filtered based on other input data, such as user email, annotations,...

I could change the behaviour of the library and make the fulcio and rekor data optional. When the user doesn't provide them the library will simply ignore (== consider not trusted) the layers that require rekor's public key and fulcio's certificate to be verified.

As for the mandatory email address that the example program requires, this is clearly a mistake. It should be made optional

What do you think about the proposal?

flavio:
I think we can close it right now, this has been merged

PR #20, opened 2022-01-12
Do not require rekor and fulcio data

flavio:
Make rekor and fulcio params not mandatory when creating a new instance of the `cosign::Client`.

This fixes issue #17


src/cosign/mod.rs
```diff
@@ -182,17 +207,28 @@ impl ClientBuilder {
...
-        let rekor_pub_key = crypto::new_verification_key(&rekor_pub_key_raw)?;
+        let rekor_pub_key = match self.rekor_pub_key {
+            None => {
+                info!("rekor public key not provided");
+                None
+            }
+            Some(raw) => Some(crypto::new_verification_key(&raw)?),
+        };
+
+        let fulcio_pub_key_raw = match self.fulcio_cert {
```
viccuad, line 225: Nit:

```suggestion
        let fulcio_pub_key_der = match self.fulcio_cert {
```

src/cosign/mod.rs
```diff
@@ -257,8 +286,14 @@ impl CosignCapabilities for Client {
     ) -> Result<Vec<SimpleSigning>> {
         let (manifest, layers) = self.fetch_manifest_and_layers(auth, cosign_image).await?;
 
-        let (_, fulcio_pub_key) = SubjectPublicKeyInfo::from_der(&self.fulcio_pub_key_raw)
-            .map_err(|e| anyhow!("Cannot parse fulcio public key: {:?}", e))?;
+        let fulcio_pub_key = match &self.fulcio_pub_key_raw {
+            None => None,
+            Some(raw) => {
```
viccuad, line 291: Nit:

```suggestion
            Some(der) => {
```

src/cosign/signature_layers.rs
```diff
@@ -217,8 +229,8 @@ impl SignatureLayer {
 pub(crate) fn build_signature_layers(
```
viccuad, line 229: The rustdoc comment of this function may not be true anymore, as the rekor and fulcio keys may be `None`. In that case, when we call `SignatureLayer::new()` on building the `signature_layers`, we are creating layers that only have the `signature` and no `bundle` or `certificate_key`, which is ok, as we can go and verify the signature. But maybe is worth noting in the rustdoc.

viccuad (approved):
LGTM modulo silly nitpicks. The untouched code has the correct precedence and outputs so verification should work as expected when the rekor and fulcio data is None, and we have tests for it.

Note, I'm still learning about the whole workflow, and just learnt about cosign bundles.

lukehinds (approved):

src/cosign/signature_layers.rs
flavio: Good catch, I fixed that now

flavio:
Thanks for all the reviews

PR #21, opened 2022-01-13
Code reorganization

flavio:
> **Note well:** this PR builds on top of https://github.com/sigstore/sigstore-rs/pull/20

The main change done by this PR is to split `cosign/lib.rs` into smaller files.
The `cosign/lib.rs` file was too big. This is now splitted into files that have a clear purpose.

This makes navigating and maintaining the code easier.

While doing that, some of the inline documentation has been moved from the specific struct into the top level library.

For example, examples about how to use `cosign::Client` have been moved into the top-level documentation.
As a result of that, examples about how to use the library are immediately shown on the main page of the library. This improves the whole developer experience.

Finally, while looking at the documentation status of the crate, I realized that `ClientBuilder::with_client_config` was a confusing name for this method.
There's a dedicated commit that renames this method to make it clear this is about the configuration used by the OCI client.


viccuad (approved):
LTGM, thanks for the quality of life improvements :). Checked that no code was changed on the move.

examples/verify/main.rs
```diff
@@ -99,15 +99,26 @@ async fn run_app() -> Result<Vec<SimpleSigning>> {
 
     let auth = &sigstore::registry::Auth::Anonymous;
 
-    let rekor_pub_key = fs::read_to_string(matches.value_of("rekor-pub-key").unwrap())
-        .map_err(|e| anyhow!("Error reading rekor public key: {:?}", e))?;
+    let rekor_pub_key: Option<String> = matches.value_of("rekor-pub-key").map(|path| {
+        fs::read_to_string(path)
+            .map_err(|e| anyhow!("Error reading rekor public key: {:?}", e))
+            .unwrap()
```
lukehinds, line 105: In time we could remove the unwraps, as read_to_string returns a Result<String>

lukehinds (approved):

lukehinds:
lgtm, happy for you to self-merge @flavio as not sure if you have a PR order to follow.

examples/verify/main.rs
lukehinds: ignore my comment, I never noticed this was example code.

PR #22, opened 2022-01-27
Add TUF support

flavio:
Allow Fulcio and Rekor data to be fetched from the official TUF repository of Sigstore.

A new module is introduced, called `tuf`, which provides helper structs to interact with a remote TUF repository.

**Note well:** this branch is using a fork of the [tough](https://github.com/awslabs/tough ) crate that includes some patches that have already been merged into the main repo.
A new release of the tough crate is going to be released in the next days: https://github.com/awslabs/tough/pull/444

This PR will be updated to use the latest official release of the tough crate.

Fixes https://github.com/sigstore/sigstore-rs/issues/9


src/tuf/repository_helper.rs
```diff
@@ -0,0 +1,345 @@
...
+use std::io::Read;
+use std::path::Path;
+use tough::{RepositoryLoader, TargetName};
+use url::Url;
+
+use super::constants::{SIGSTORE_FULCIO_CERT_TARGET, SIGSTORE_REKOR_PUB_KEY_TARGET};
+
+pub(crate) struct RepositoryHelper {
+    repository: tough::Repository,
+    checkout_dir: Option<String>,
```
ereslibre, line 28: I would favour a `PathBuf` instead of a `String`. It helps to avoid multi-platform gotchas and more accurate to the intent of `checkout_dir`.

src/tuf/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+    ///    .await
+    ///    .expect("Error spawning blocking task");
+    ///
+    ///    // handle the case of `repo` being an `Err`
+    ///    // ... your code
+    /// }
+    /// ```
+    ///
+    /// This of course has a performance hit when used inside of an async function.
+    pub fn fetch(checkout_dir: Option<&str>) -> Result<Self> {
```
ereslibre, line 121: I think using `Option<&Path>` or `Option<PathBuf>` is a better API and closer to the user intent.

src/tuf/repository_helper.rs
```diff
@@ -0,0 +1,345 @@
...
+/// didn't change.
+/// This check is done by comparing the digest of the local file, if found,
+/// with the digest reported inside of the TUF repository metadata.
+///
+/// **Note well:** the `local_file` is updated whenever its contents are
+/// outdated.
+fn fetch_target_or_reuse_local_cache(
+    repository: &tough::Repository,
+    target_name: &TargetName,
+    local_file: Option<&std::path::PathBuf>,
```
ereslibre, line 108: Given that we are having a ref already as the associated type of the `Option`, I think it would be good to switch to `Option<&Path>` as the argument.

This would allow a caller to call with a `&Path`, or with a `&PathBuf`, because `PathBuf` implements `Deref<Target = Path>`: https://doc.rust-lang.org/std/path/struct.PathBuf.html#associatedtype.Target.

ereslibre (commented):
Thanks @flavio.

A couple of minor comments.

src/tuf/repository_helper.rs
flavio: Good point, fixed

src/tuf/mod.rs
flavio: Fixed

src/tuf/repository_helper.rs
flavio: Fixed

ereslibre (approved):
LGTM. Approving from my point of view although I'm not a maintainer.

flavio:
Thanks!

The tough maintainers have just tagged a new release with all the fixes we need :partying_face: 

I've updated the PR to use this new revision

PR #24, opened 2022-01-31
Better error handling

flavio:
Do not rely on `anyhow::Result` inside of public APIs. This makes error handling uglier for consumers of the library.

Instead, provide custom errors and use them inside of all the public facing APIs.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,0 +1,102 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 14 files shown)


src/errors.rs
```diff
@@ -0,0 +1,102 @@
...
+    #[error(transparent)]
+    Base64DecodeError(#[from] base64::DecodeError),
+
+    #[error(transparent)]
+    EcdsaError(#[from] ecdsa::Error),
+
+    #[error("Certificate validity check failed: cannot be used before {0}")]
+    CertificateValidityError(String),
+
+    #[error("Certificate has not been issued for the {0}")]
```
viccuad, line 43: Nitpick:

```suggestion
    #[error("Certificate has not been issued for {0}")]
```
Just so it prints `Certificate has not been issued for foo@bar.com`.

viccuad (approved):
LGTM! Cool crate, nice cleanup, and we got a couple of new error types and prints too.

lukehinds (approved):

PR #26, opened 2022-02-02
Add functionality to make an entry in Rekor

jyotsna-penumaka:
This is a draft pull request, but I would really appreciate getting some feedback.

Signed-off-by: jyotsna-penumaka <jpenumak@redhat.com>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,13 +21,17 @@ oci-distribution = { version = "0.8.1", default-features = false }
 olpc-cjson = "0.1.1"
 p256 = {version = "0.9.0", features = ["ecdsa-core"]}
 sha2 = "0.10.1"
-serde_json = "1.0.68"
-serde = {version = "1.0.130", features = ["derive"]}
-tokio = { version = "1.12.0", features = ["full"]}
+serde_json = "1.0.78"
+serde = {version = "1.0.136", features = ["derive"]}
+tokio = { version = "1.16.1", features = ["full"]}
 tough = { version = "0.12.1", features = [ "http" ] }
 tracing = "0.1.29"
 url = "2.2.2"
 x509-parser = { version = "0.12.0", features = ["verify"]}
+
+serde_derive = "1.0.136"
+reqwest = { version = "0.11.9", features = ["json"] }
+openssl-probe = "0.1.5"
 
 [dev-dependencies]
 chrono = "0.4.19"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,6 @@ x509-parser = { version = "0.12.0", features = ["verify"]}
 
 serde_derive = "1.0.136"
 reqwest = { version = "0.11.9", features = ["json"] }
-openssl-probe = "0.1.5"
-httpmock = "0.6"
 
 [dev-dependencies]
 anyhow = "1.0.44"
@@ -41,3 +39,4 @@ clap = "2.33.3"
 openssl = "0.10.38"
 tempfile = "3.3.0"
 tracing-subscriber = "0.2.25"
+httpmock = "0.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ x509-parser = { version = "0.12.0", features = ["verify"]}
 serde_derive = "1.0.136"
 reqwest = { version = "0.11.9", features = ["json"] }
 openssl-probe = "0.1.5"
+httpmock = "0.6"
 
 [dev-dependencies]
 anyhow = "1.0.44"
--- a/src/rekor.rs
+++ b/src/rekor.rs
@@ -45,7 +45,7 @@ pub struct Signature {
     pub public_key: PublicKey,
 }
 
-/// Stores the public key used to sign the artifact 
+/// Stores the public key used to sign the artifact
 #[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct PublicKey {
@@ -68,7 +68,7 @@ pub struct Hash {
     pub value: String,
 }
 
-/// Stores the response returned by Rekor after making a new entry 
+/// Stores the response returned by Rekor after making a new entry
 #[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Post {
@@ -98,7 +98,6 @@ pub async fn rekor_upload(
     algorithm_name: String,
     hash_val: String,
 ) -> Result<Post, Box<dyn Error>> {
-
     // This is the body of the post request
     let body = Root {
         api_version: api_version_val,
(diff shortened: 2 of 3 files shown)


jyotsna-penumaka:
@lkatalin Thanks for the review.
Since we can only make a unique rekor entry, the test that is currently in the file passed (the request to make a rekor entry would return a Post struct) the first time (with fresh data - new file's hash and signature) but now it would fail (it would return an error). To test the code, you will have to use another file and its hash and signature. Do you have any recommendations for the tests to add?
I don't think there is more work to be done for the functionality of making a rekor entry.
I could add the functionality to get a rekor entry and the other cli commands in another pull request.

lkatalin:
> To test the code, you will have to use another file and its hash and signature. Do you have any recommendations for the tests to add?
I see why that's tricky. You would have to generate some new test data for each call to the test, instead of hardcoding it. You could have some function in the tests module `generate_test_data()` that echoes a randomly generated string to a file, signs that file ex. with `ssh-keygen`, and returns the path to the file and its signature file. Then you could upload those to Rekor, and remove the files you generated. Generating the data would almost make more sense to do in bash, though. Or maybe we don't need to get so complicated. @lukehinds Do you have any thoughts here? Is it enough to have Rekor say "this entry already exists" or do we want to fully test that we're uploading new entries?

> I could add the functionality to get a rekor entry and the other cli commands in another pull request.
Sounds good!

src/rekor.rs
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+    */
+
+    fn get_github_token() -> String {
+        match env::var("GITHUB_AUTH_TOKEN") {
+            Ok(val) => format!("Token {}", val),
+            Err(_err) => ("").to_string(),
+        }
+    }
```
lukehinds, line 87: I don't think we need this?

src/rekor.rs
```diff
@@ -0,0 +1,166 @@
...
+        let new_post = reqwest::Client::new()
+            .post("https://rekor.sigstore.dev/api/v1/log/entries")
+            .header("Authorization", get_github_token())
+            .json(&new_post)
+            .send()
+            .await
+            .unwrap()
+            .text()
+            .await
+            .unwrap();
```
lukehinds, line 129: I will take another look on Monday, but we might be able to trim these unwraps with a question mark operator

lukehinds (commented):
A couple of comments made. One thing that might be useful is an entry in the examples folder.

lukehinds:
> > To test the code, you will have to use another file and its hash and signature. Do you have any recommendations for the tests to add?
> 
> I see why that's tricky. You would have to generate some new test data for each call to the test, instead of hardcoding it. You could have some function in the tests module `generate_test_data()` that echoes a randomly generated string to a file, signs that file ex. with `ssh-keygen`, and returns the path to the file and its signature file. Then you could upload those to Rekor, and remove the files you generated. Generating the data would almost make more sense to do in bash, though. Or maybe we don't need to get so complicated. @lukehinds Do you have any thoughts here? Is it enough to have Rekor say "this entry already exists" or do we want to fully test that we're uploading new entries?
> 
> > I could add the functionality to get a rekor entry and the other cli commands in another pull request.
> 
> Sounds good!

Ideally we should test for both new entry and existing entry. Maybe I have not fully grasped the issues here, but we could have to sets of signing materials in a json file, that we parse for both tests?

src/rekor.rs
flavio: I agree, I don't think sigstore-rs should provide this helper. IMHO it's up to the developers using the library to provide their token (which could be obtained by them, inside of their code, using something like this code you wrote).

src/rekor.rs
```diff
@@ -0,0 +1,166 @@
+#[cfg(test)]
+mod tests {
+    use serde_derive::Deserialize;
+    use serde_derive::Serialize;
+    use std::env;
+
+    #[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Root {
```
flavio, line 9: I think it would make sense to provide a custom implementation of the `Default` trait, so that certain types have a meaningful value. For example, `api_version` and `kind` have probably something well known.

If that's the case, these values should be defined via constants

src/rekor.rs
```diff
@@ -0,0 +1,166 @@
...
+    async fn rekor_upload(
+        api_version_val: String,
+        object_type: String,
+        key_format: String,
+        signature_val: String,
+        pub_key_val: String,
+        file_url: String,
+        algorithm_name: String,
+        hash_val: String,
+    ) -> Result<Post, serde_json::Error> {
```
flavio, line 98: This is a matter of personal taste, I don't like functions with many parameters to be exposed to the end users. This can get confusing.

What about using the [builder pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) (see the cosign::ClientBuilder as another example).

We could have a `RootBuilder` struct, that builds `Root` objects based on a mix of default values and the values provided by the user.

Then we could have a `RekorClient` struct which has a `submit(root: &Root) -> crate::errorsResult<Post>` method.

The `RekorClient` could then take care of creating the reqwest client using the right set of attributes. Right now we will probably use the default params (like you're doing), but this architecture would allow us - if needed - to fine tune the rekor client to do things like:
  * talk with a different Rekor server - for people running their infra (this could be useful also for end to end tests)
  * load additional trusted CAs and do other tuning required to talk with "self-hosted" Rekor server

Most important of all, we can have a mock reqwest client to avoid direct interactions with the real Rekor server during our unit tests.

Cargo.toml
```diff
@@ -20,15 +20,19 @@ oci-distribution = { version = "0.8.1", default-features = false }
...
 thiserror = "1.0.30"
-tokio = { version = "1.12.0", features = ["full"]}
 tough = { version = "0.12.1", features = [ "http" ] }
 tracing = "0.1.29"
 url = "2.2.2"
 x509-parser = { version = "0.12.0", features = ["verify"]}
 
+serde_derive = "1.0.136"
+reqwest = { version = "0.11.9", features = ["json"] }
+openssl-probe = "0.1.5"
```
flavio, line 34: I don't think we need this crate, certificates are already found by openssl-sys and rustls.

src/rekor.rs
```diff
@@ -0,0 +1,166 @@
...
+            .send()
+            .await
+            .unwrap()
+            .text()
+            .await
+            .unwrap();
+
+        println!();
+        println!("{:#?}", new_post);
+        println!();
```
flavio, line 133: Do not use `println`, use `tracing` like inside of other parts of the library. When producing traces please use structured logging too.

src/rekor.rs
```diff
@@ -0,0 +1,166 @@
...
+        } else {
+            println!("There is an error! Cannot parse the response :( ");
+        }
+        println!();
+
+        let post = new_post;
+        let uuid: &str = &post[1..67];
+        let rest: &str = &post[69..post.len() - 2];
+        let sum = "{\"uuid\": ".to_string() + &(uuid.to_owned()) + "," + rest;
+        let v: Result<Post, serde_json::Error> = serde_json::from_str(&sum);
```
flavio, line 146: I'm a bit lost, are you parsing the json response manually?

src/rekor.rs
```diff
@@ -0,0 +1,166 @@
...
+    #[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Hash {
+        pub algorithm: String,
+        pub value: String,
+    }
+
+    #[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct Post {
```
flavio, line 52: I think a better name for that could be `LogEntry`. The name is based on what rekor uses internally, unless I got it wrong by looking inside of rekor's [code](https://github.com/sigstore/rekor/blob/main/pkg/generated/models/log_entry.go#L36)

src/rekor.rs
flavio: Now that I got to the end of the PR, and looked at the rekor source code, this seems to be the most generic [ProposedEntry](https://github.com/sigstore/rekor/blob/3c0d61f848b7e22277f1cefb8d15c896917fdcf3/pkg/api/entries.go#L147).

Should we rename `Root` to be `ProposedEntry`? Maybe in the future, not right now, we could have a trait and then have one struct per type that we support, something like [rekor does](https://github.com/sigstore/rekor/blob/main/pkg/types/README.md)

src/rekor.rs
```diff
@@ -0,0 +1,166 @@
...
+        let uuid: &str = &post[1..67];
+        let rest: &str = &post[69..post.len() - 2];
+        let sum = "{\"uuid\": ".to_string() + &(uuid.to_owned()) + "," + rest;
+        let v: Result<Post, serde_json::Error> = serde_json::from_str(&sum);
+        v
+    }
+
+    #[tokio::test]
+    async fn verify_rekor_upload() -> Result<(), reqwest::Error> {
+        let _new_post = rekor_upload(
```
flavio, line 152: As said before, I would not hit the real rekor server inside of the unit tests. We should instead use a mock to simulate it.

flavio (changes requested):
I feel sorry about the number of comments I left, I hope you won't mind :sweat_smile:

src/rekor.rs
lkatalin: > As said before, I would not hit the real rekor server inside of the unit tests. We should instead use a mock to simulate it.

This is a good suggestion, and it should let us avoid having to generate new test data for each time we run the unit tests. So @jyotsna-penumaka, please ignore my previous comment about generating the data.

lukehinds: ah I just groked while you @ mentioned me now about dupe entries, I did not spot this was against a live server. +1 to mocking

Cargo.toml
```diff
@@ -20,19 +20,24 @@ oci-distribution = { version = "0.8.1", default-features = false }
...
+httpmock = "0.6"
```
lukehinds, line 35: we should be able to move this under dev-dependencies.

jyotsna-penumaka: I added it to dev-dependencies now.
I moved the rekor_upload function that creates the mock server and send the request inside mod tests.

src/rekor.rs
```diff
@@ -0,0 +1,166 @@
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
jyotsna-penumaka: Thank you for the advice!
`cargo doc --open` is a very nice feature.
I added documentation for the functions and the structs.

src/rekor.rs
jyotsna-penumaka: I thought renaming `Root` to `Rekord` would make more sense, since this request body corresponds to a Rekord entry.
I used the builder pattern like you suggested and initialized `api_version`, `kind` in `Rekord` and `format` in `Signature` and `algorithm` in `Hash` structs with the default values.

src/rekor.rs
jyotsna-penumaka: Yeah I completely agree, builder pattern looks much cleaner and is easier to understand :)

I used the builder pattern for the Rekord struct but I am not very clear on the RekorClient.
Is the idea to create a RekorClient struct, where a user can specify the server they want to make the request to?

src/rekor.rs
jyotsna-penumaka: We need to modify the response before we can read it into a Struct
This is the response that is returned by Rekor:

        {
            "9bf4f37447a48848c1f69b6463190cd6cef7728d6b86b7723f08aad0c54cb8d4": {
                "body": "eyJhcGlWZXJzaW9uIjoiMC4wLjEiLCJraW5kIjoicmVrb3JkIiwic3BlYyI6eyJkYXRhIjp7Imhhc2giOnsiYWxnb3JpdGhtIjoic2hhMjU2IiwidmFsdWUiOiI0NmJkMzE5ZDM1OTBkMmI0ZDdjN2EyN2M5OWQzMmY3ZWE2MGE4NTBlYzM4MDYzNTFlMDRkMTYxZDAxNGVjYzAxIn19LCJzaWduYXR1cmUiOnsiY29udGVudCI6IkxTMHRMUzFDUlVkSlRpQlRVMGdnVTBsSFRrRlVWVkpGTFMwdExTMEtWVEZPU1ZVd2JFaEJRVUZCUVZGQlFVRkVUVUZCUVVGTVl6Tk9iMHhYVm10TmFsVXhUVlJyUVVGQlFXY3ZkbVZUWXpSdmJIQkxkRTF2VDFJM2NuZG1PQXBXUjBoNmFHaG5NRVpKYjBSMFl6VlNNa3BzZEhwSFowRkJRVUZGV20xc2MxcFJRVUZCUVVGQlFVRkJSMk15YUdoT1ZFVjVRVUZCUVZWM1FVRkJRWFI2Q21NeVozUmFWMUY1VGxSVmVFOVJRVUZCUlVKalEydDBaME14V1dwcmIzZEtkSEJzZVhCRFNEUTJhbkV5UW1Sb05tUjZhblIwZVd0SFpWRjVLMG8xZUhvS2RIRlVlbXRqUXpCWVNHRkZZWEZPZFdNMWNURnpUbEZNWTJRNFNEUjRNM0ZLU2xSRFFsRnZUd290TFMwdExVVk9SQ0JUVTBnZ1UwbEhUa0ZVVlZKRkxTMHRMUzBLIiwiZm9ybWF0Ijoic3NoIiwicHVibGljS2V5Ijp7ImNvbnRlbnQiOiJjM05vTFdWa01qVTFNVGtnUVVGQlFVTXpUbnBoUXpGc1drUkpNVTVVUlRWQlFVRkJTVkEzTTJ0dVQwdEtZVk55VkV0RWEyVTJPRWd2UmxKb09EUlpXVTVDVTB0Qk4xaFBWV1JwV21KamVHOEsifX19fQ==",
                "integratedTime": 1643749079,
                "logID": "c0d23d6ad406973f9559f3ba2d1ca01f84147d8ffc5b8445c224f98b9591801d",
                "logIndex": 1236200,
                "verification": {
                    "signedEntryTimestamp": "MEQCID3QHHXwGauKUfFvs0YCMKZ4e3BE1ZIhrJ5RHnoHGophAiA/R5cDc8JnSUq6F7U/8oTk8/mn7QCdDXI+NRT7qm78+Q=="
                }
            }
        }
        
Since there is no member called "uuid" in the json body,
we cannot read it into the LogEntry struct.
So we add "{\"uuid\": " to the returned response and make it a valid json

src/rekor.rs
jyotsna-penumaka: I think the CI clippy test seems to run into an error when the new function returns the builder and now self.

jyotsna-penumaka:
Thank you @flavio , @lukehinds and @lkatalin for the detailed suggestions.
They were very helpful!

lukehinds:
shall we close this now that you have your crate @jyotsna-penumaka ?

lukehinds:
To add though, we should have something in examples/ or even a light client implementation.

jyotsna-penumaka:
@lukehinds I agree, I think we should close this.
I will add the examples to this repo once I publish the crate.

Issue #18, opened 2021-12-21
Fail on verify() if image has no signatures


viccuad:
**Description**

Right now, if one uses `verify()` against an image that doesn't contain signatures, it will return `Ok()`, with an empty `Vec<SimpleSigning>`
See: https://github.com/sigstore/sigstore-rs/blob/cc9fe808918f2b5775143e8ade351d315044db30/src/cosign/mod.rs#L276-L282

As a user, I want to be sure that unsigned images don't slip through. Hence, I would prefer if `verify()` returns an Err when there are no signatures to be verified.

This way, I can delegate the responsibility to the library, and I don't need to check on my own if an image has signatures prior to calling verify() on it.

viccuad:
Closed by https://github.com/sigstore/sigstore-rs/pull/31! :partying_face:

PR #31, opened 2022-02-07
 Big code refactoring: better handling of verification

flavio:
First of all, sorry about the large amount of changes... Unfortunately the old code was too limited as we realized when using that inside of Kubewarden.

Prior to this commit:

* The email used to verify a keyless signature had to be specified inside of the `ClientBuilder`. That meant that, to verify the presence of multiple emails, multiple clients had to be created. That was a poor UX
* The library didn't parse some important data provided by the signature certificates: the URI and the OIDC issuer. These are required to build more flexible verification checks

In addition to addressing the previous limitations, this commit introduces the following changes:

* It's possible to define a series of Verification Constraints, which allows multiple requirements can be enforced against the list of signature layers of an image
* End users can develop their own Verification Constraints, and feed them to the usual verification API to enforce them
* Not finding any trusted layer now raises a dedicated error. This fixes issue #18

### DIFF
--- a/examples/verify/main.rs
+++ b/examples/verify/main.rs
@@ -14,9 +14,14 @@
 // limitations under the License.
 
 extern crate sigstore;
+use sigstore::cosign::verification_constraint::{
+    AnnotationVerifier, CertSubjectEmailVerifier, CertSubjectUrlVerifier, PublicKeyVerifier,
+    VerificationConstraintVec,
+};
 use sigstore::cosign::CosignCapabilities;
-use sigstore::simple_signing::SimpleSigning;
+use sigstore::cosign::SignatureLayer;
 use sigstore::tuf::SigstoreRepository;
+use std::boxed::Box;
 
 extern crate anyhow;
 use anyhow::anyhow;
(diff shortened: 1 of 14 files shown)


examples/verify/main.rs
```diff
@@ -104,9 +129,15 @@ fn cli() -> App<'static, 'static> {
         )
 }
 
-async fn run_app() -> anyhow::Result<Vec<SimpleSigning>> {
+async fn run_app() -> anyhow::Result<Vec<SignatureLayer>> {
     let matches = cli().get_matches();
 
+    if matches.is_present("cert-email") && matches.is_present("cert-url") {
+        return Err(anyhow!(
```
flavio, line 136: This is more of a limitation of this example program, not of sigstore. Maybe I should add a comment about that?

examples/verify/main.rs
```diff
@@ -167,67 +198,79 @@ async fn run_app() -> anyhow::Result<Vec<SimpleSigning>> {
...
-        .build()?;
-
-    let image: &str = matches.value_of("IMAGE").unwrap();
+    let mut client = client_builder.build()?;
 
-    let (cosign_signature_image, source_image_digest) = client.triangulate(image, auth).await?;
+    // Build verification constraints
+    let mut verification_constraint: VerificationConstraintVec = Vec::new();
+    if let Some(cert_email) = matches.value_of("cert-email") {
+        let issuer = matches.value_of("cert-issuer").map(|i| i.to_string());
```
lukehinds, line 206: do we not need to check if issuer is also not `is_none` here?

flavio: Not in this case, the `CertSubjectEmailVerifier` has an `Option<String>` for issuer. Hence it's fine if this is `None`.

The main reasoning behind that is: assuming the email is `alice@gmail.com`, I might not care if Alice authenticated herself via GitHub (where she uses her gmail account) or directly via Google.

There's also the possibility to specify an issuer, which would then ensure both the email and the issuer of the verifier match with the data inside of the certificate attached to a signature

lukehinds (approved):

examples/verify/main.rs
flavio: I've added a comment about that

flavio:
The patches I've submitted to oci-distribution have been merged into the `main` branch. I've updated the `Cargo.toml` to build from a specific revision of `krustlet/oci-distribution` instead of my own personal fork.

I've also bumped the minor version because of the API changes this PR introduces.

lukehinds:
> The patches I've submitted to oci-distribution have been merged into the `main` branch. I've updated the `Cargo.toml` to build from a specific revision of `krustlet/oci-distribution` instead of my own personal fork.
> 
> I've also bumped the minor version because of the API changes this PR introduces.

awesome, thanks for all your great work @flavio

PR #35, opened 2022-02-11
feat: Verify annotations in key, email, url verifiers 

viccuad:
#### Summary

For PublicKeyVerifier, CertSubjectEmailVerifier, CertSubjectUrlVerifier,
add a new `annotations` parameter, and logic so they check that their
layer contains the passed annotations.

This way, we can for example create a PublicKeyVerifier with a key and
annotations. When filtering with cosign::filter_signature_layers(), we
will filter for those signatures signed with that key and containing
those annotations.

This is in contrast with the already existing AnnotationsVerifier, that
is used to filter for those signatures containing the
annotations (regardless of the identity), and it's used to check that
all signatures contain the passed annotations.

Consume changes to verifiers.

This is useful in github.com/kubewarden/policy-fetcher, as we verify signatures with annotations,
apart from annotations globally for all signatures.

#### Release Note

```release-note
feat: Added a new `annotations` parameter to `PublicKeyVerifier`, `CertSubjectEmailVerifier`, `CertSubjectUrlVerifier`.
```

### DIFF
--- a/examples/verify/README.md
+++ b/examples/verify/README.md
@@ -1,6 +1,6 @@
-This is a simple example program that shows how perform cosign verification.
+This is a simple example program that shows how to perform cosign verifications.
 
-The program allows also to use annotation, in the same way as `cosign verify -a key=value`
+The program allows also to use annotations, in the same way as `cosign verify -a key=value`
 does.
 
 The program prints to the standard output all the Simple Signing objects that
--- a/examples/verify/main.rs
+++ b/examples/verify/main.rs
@@ -202,6 +202,19 @@ async fn run_app() -> anyhow::Result<Vec<SignatureLayer>> {
     let mut client = client_builder.build()?;
 
     // Build verification constraints
+    let annotations = if let Some(annotations) = matches.values_of("annotations") {
+        let mut values: HashMap<String, String> = HashMap::new();
+        for annotation in annotations {
+            let tmp: Vec<_> = annotation.splitn(2, "=").collect();
+            if tmp.len() == 2 {
+                values.insert(String::from(tmp[0]), String::from(tmp[1]));
+            }
+        }
+        values
+    } else {
+        HashMap::default()
+    };
+
     let mut verification_constraint: VerificationConstraintVec = Vec::new();
     if let Some(cert_email) = matches.value_of("cert-email") {
         let issuer = matches.value_of("cert-issuer").map(|i| i.to_string());
@@ -209,6 +222,7 @@ async fn run_app() -> anyhow::Result<Vec<SignatureLayer>> {
         verification_constraint.push(Box::new(CertSubjectEmailVerifier {
             email: cert_email.to_string(),
             issuer,
+            annotations: annotations.clone(),
         }));
     }
     if let Some(cert_url) = matches.value_of("cert-url") {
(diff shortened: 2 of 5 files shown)


viccuad:
Force-pushed for Co-author signing off.

flavio:
I have some doubts about these changes. I think the addition of `annotations` to each verifier is "cluttering" the API.

From my understanding, this is your use case. Given a public key and a list of annotations, find all the signatures produced for container image `<foo>` that have been produced by this public key AND satisfy all the specified annotations.

This can already be done via the combined usage of  of `PublicKeyVerifier` and `AnnotationVerifier` inside of the `sigstore::cosign::filter_signature_layers` function.

You have to create a `VerificationConstraintVec` instance, add to it the two constraints and then pass it to the `filter_signature_layers`.
This will returns a list of layers that are satisfying both your constraints.

The current code is less optimized compared to the one you're proposing. But I don't think we will have thousand of signature layers for each signed artifact, plus hundreds of verification constraints. If that happens, the end users of the library can still create their own verifiers that are doing this sort of optimizations.

I would prefer the keep the code as it is, it feels more "clean" to me.

flavio:
I'm now reviewing the PR you created on Kubewarden that leverages the new verification API. I now understand why you came up with this PR.

Kubewarden has a slightly different use case than the one I described.

Some quick background, this is a possible verification configuration:

```yaml
- kind: pubKey
  owner: flavio # not relevant - just to understand who owns this key
  key: .... # mandatory
  annotations: # optional
    env: prod
- kind: pubKey
  owner: alice # not relevant - just to understand who owns this key
  key: .... # mandatory
  annotations: # optional
    env: prod
    reviewed: true
```

In this case, all the layers signed by `flavio` must have the `env=prod` annotation, while the ones signed by `alice` must have `env=prod AND reviewed=true`.
That's why the current verifier are sub-optimal to this use case.

However, I think the API of sigstore-rs should be as generic as possible, and just allow end consumers to provide their own implementations when needed.

In this specific example, I would declare  the following verifier inside of Kubewarden:

```rust
#[derive(Debug)]
struct PubKeyAndAnnotationsVerifier {
    pub_key_verifier: PubKeyVerifier,
    annotation_verifier: AnnotationVerifier,
}

impl PubKeyAndAnnotationsVerifier {
    pub fn new(key: &str, annotations: HashMap<String, String>) -> Result<Self> {
        let pub_key_verifier = PubKeyVerifier::new(key)?;
        let annotation_verifier = AnnotationVerifier { annotations };
        Ok(Self {
            pub_key_verifier,
            annotation_verifier,
        })
    }
}

impl VerificationConstraint for PubKeyAndAnnotationsVerifier {
    fn verify(&self, sl: &SignatureLayer) -> Result<bool> {
        let outcome = self.annotation_verifier(sl)? && self.pub_key_verifier(sl)?;
        Ok(outcome)
    }
}
```

A similar one could be done to compose the annotation and the cert verifiers together.

These new verifiers would then be used by Kubewarden and provided to the filter function.

I think this would be the right compromise between a generic API for sigstore-rs, and code-reuse inside of Kubewarden.

What do you think?

viccuad:
That's a good proposal, I am happy composing Verifiers in Kubewarden instead. It simplifies things, and we take a performance hit from O(n) to O(n*m), n being present layers and m being number of signatures to verify, but yes, the differing signatures are going to be ~10 tops.

Closing in favor of that.

PR #36, opened 2022-02-11
Create GH action to auto publish crate upon release

### LINKED ISSUE #30: Automate crates publish of releases
Perhaps we should utilise a GH action too `cargo publish` each release?

https://github.com/marketplace/actions/publish-crates

cc @flavio @cpanato

cburkhardt27:
#### Summary

Creates a Github action to automatically `cargo publish` each new release to crates.io.
The version number of the crate in the Cargo.toml file will have to be updated prior to the new release for the action to work.
Also, a Github secret for the crates.io API token needs to be set up so that there's authorization to publish the update. Currently, the secret is named CARGO_API_TOKEN in the .yml file.
#### Ticket Link

Fixes #30 

#### Release Note

```release-note

```

### DIFF
--- a/.github/workflows/auto-publish-crates-upon-release.yml
+++ b/.github/workflows/auto-publish-crates-upon-release.yml
@@ -14,6 +14,4 @@ jobs:
            override: true
 
       - name: publish crates
-        uses: katyo/publish-crates@v1
-        with: 
-          registry-token: ${{ secrets.CARGO_API_TOKEN }}
+        run: cargo publish --token ${{ secrets.CARGO_API_TOKEN }}
--- a/.github/workflows/auto-publish-crates-upon-release.yml
+++ b/.github/workflows/auto-publish-crates-upon-release.yml
@@ -1,0 +1,19 @@
+name: Publish-Crate-Upon-Release
+on:
+  release:
+    types: [published]
+
+jobs:
+  publish-automatically:
+    runs-on: ubuntu-latest
+    steps:
+      -  uses: actions/checkout@v2
+      -  uses: actions-rs/toolchain@v1
+         with:
+           toolchain: stable
+           override: true
+
+      - name: publish crates
+        uses: katyo/publish-crates@v1
+        with: 
+          registry-token: ${{ secrets.CARGO_API_TOKEN }}


.github/workflows/auto-publish-crates-upon-release.yml
```diff
@@ -0,0 +1,19 @@
...
+      -  uses: actions/checkout@v2
+      -  uses: actions-rs/toolchain@v1
+         with:
+           toolchain: stable
+           override: true
+
+      - name: publish crates
+        uses: katyo/publish-crates@v1
+        with: 
+          registry-token: ${{ secrets.CARGO_API_TOKEN }}
```
flavio, line 19: This is a matter of personal taste, but I'm a bit "paranoid" about depending on external GitHub actions. Especially if they don't have much adoption or if they are doing something "trivial".

I've looked at the features provided by this extension and I couldn't find something that we need and is not covered by `cargo publish`. IMHO `cargo publish` is already doing the basic checks we would need.

I would start with something like that, and then eventually migrate to this action or to something else if we need more features:

```suggestion
      - name: publish crates
        run: cargo publish --token ${{ secrets.CARGO_API_TOKEN }}
```

lukehinds: That' a fair point. We also need to think about whose token we use, as its tied to a GitHub oauth2 grant. I am happy to self-nominate my account, but I also think we should include other sigstore-rs codeowners (aka yourself @flavio)

lukehinds: I added a token for now under `CARGO_API_TOKEN`, we can try that out @cburkhardt27

flavio: A common pattern I've seen on crates.io is the following one:

  * Create a bot account on crates-io: add it as one of the owners of the crate
  * Add the human maintainers of the library as maintainers too

We might want to do that... I don't know if the Sigstore organization has already a bot account that is used for other purposes. Otherwise we could simply create a new bot account on crates.io

lukehinds: OK. We now have a sigstore-bot account on crates.io: https://crates.io/users/sigstore-bot

The bot's token is added to `CARGO_API_TOKEN`

@cburkhardt27 , you should be able to make the changes flavio recommends.

@flavio Let us know when the oci-distribution PR merges that we need before we can make the first release.

cburkhardt27: Thanks for the feedback! I updated the PR to `cargo publish`

flavio (approved):
LGTM

lukehinds (approved):
Thanks @cburkhardt27 !

PR #38, opened 2022-02-21
Change the visibility of `SignagureLayer` attributes

flavio:
This is required to make writing of unit tests easier for 3rd parties defining their own constraint validators.


viccuad (approved):
Don't have review rights here, yet thanks!

lukehinds (approved):

PR #40, opened 2022-02-24
fix: cosign::filter_signature_layers() when no match 

viccuad:
#### Summary

Remove negation inside of the `match` for constraints. This means that verified layers will get correctly passed to `any()`.

Remove negation on the `constraints.iter.any()`. this means that when a layer is verified (returning true in the closure), it well get added correctly to the layers vector.

All in all, this gives us the correct behaviour: if there's some verified signatures, they get added.
If there's no verified signatures, we return `SigstoreError::SigstoreNoVerifiedLayer`. In the past, we where almost always returning that error instead.

Fix test for correct behaviour.

#### Release Note

```release-note
- fx: `cosign::filter_signature_layers()` when no match
```

### DIFF
--- a/src/cosign/mod.rs
+++ b/src/cosign/mod.rs
@@ -275,10 +275,7 @@ Hr/+CxFvaJWmpYqNkLDGRU+9orzh5hI2RrcuaQ==
 
         let error =
             filter_signature_layers(&layers, constraints).expect_err("Should have got an error");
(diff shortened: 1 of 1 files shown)


viccuad:
Rebased onto main.

flavio (approved):
Thanks for the fix and for the updated test case

Issue #41, opened 2022-03-02
`filter_signature_layers()` is insufficient to verify images, and may provide false positives


viccuad:
**Description**

Current [`sigstore::cosign::filter_signature_layers()`](https://github.com/sigstore/sigstore-rs/blob/9b561afb03cbd21f5960cf5628e7e99354f5e3c2/src/cosign/mod.rs#L94-L125) returns a list of signatures passing the constraints. This is not enough to verify if all the constraints have been satisfied; a constraint could make several layers pass at once, as applying constraints is a non-injective function (for example filtering with `AnnotationVerifier` will give us several layers that have passed verification).
For the other verifiers, we didn't realize, as they verify certs and pub keys, and normally there aren't duplicated signatureLayers with those.

We don't gain much information from `filter_signature_layers()` returning a vector of verified layers. We can calculate the unverified layers from that vector, but none of those 2 vectors give us information on satisfied or unsatisfied constraints.

**Solution**

Change `cosign::filter_signature_layers()` to return unsatisfied constraints.

What we really care about is filtering the constraints into those that got satisfied, and those that not. Consumers then can know that an image failed verification if there are unsatisfied constraints, or the image passed verification if all constraints were satisfied.

Also, information on *why* an image didn't pass verification can only be obtained from the list of unsatisfied constraints. I expect consumers to have verifiers that implement  `VerificationConstraint` and `std::fmt:Display`, for example, to print to users those unsatisfied constraints.

viccuad:
I have started to implement this here.

PR #42, opened 2022-03-04
fix!: Drop filter_signature_layers() for verify_constraints() 

viccuad:
#### Summary

BREAKING CHANGE: Drop `cosign::filter_signature_layers` in favour of
`cosign::verify_constraints()`, that returns a new `SigstoreVerifyConstraintsError`
which contains references to unsatisfied constraints.

The relationship between layers and constraints is not injective (e.g:
AnnotationsVerifier), hence no image verification info could be derived
from the number of layers verified.

With unsatisfied constraints, we know that 0 means the image is
verified.
If we have some unsatisfied constraints, we can use them to provide info
on why we failed verification.

Add a test and consume changes.

#### Ticket Link

Fixes https://github.com/sigstore/sigstore-rs/issues/41.

#### Release Note

```release-note
fix!: Drop filter_signature_layers() for verify_constraints()
```

### DIFF
--- a/examples/verify/main.rs
+++ b/examples/verify/main.rs
@@ -16,7 +16,7 @@
 extern crate sigstore;
 use sigstore::cosign::verification_constraint::{
     AnnotationVerifier, CertSubjectEmailVerifier, CertSubjectUrlVerifier, PublicKeyVerifier,
-    VerificationConstraint, VerificationConstraintVec,
+    VerificationConstraintVec,
 };
 use sigstore::cosign::{CosignCapabilities, SignatureLayer};
 use sigstore::crypto::SignatureDigestAlgorithm;
@@ -93,7 +93,7 @@ struct Cli {
     image: String,
 }
 
-async fn run_app() -> anyhow::Result<(Vec<SignatureLayer>, Vec<Box<dyn VerificationConstraint>>)> {
+async fn run_app() -> anyhow::Result<(Vec<SignatureLayer>, VerificationConstraintVec)> {
     let cli = Cli::parse();
 
     // Note well: this a limitation deliberately introduced by this example.
--- a/examples/verify/main.rs
+++ b/examples/verify/main.rs
@@ -232,7 +232,7 @@ pub async fn main() {
                 0
             } else {
                 eprintln!("Image verification failed: not all constraints satisfied.");
-                // serde_json::to_writer_pretty(std::io::stdout(), &unsatisfied_constraints).unwrap();
+                println!("{:?}", unsatisfied_constraints);
                 1
             }
         }
--- a/examples/verify/main.rs
+++ b/examples/verify/main.rs
@@ -19,7 +19,6 @@ use sigstore::cosign::verification_constraint::{
     VerificationConstraintVec,
 };
 use sigstore::cosign::CosignCapabilities;
-use sigstore::cosign::SignatureLayer;
 use sigstore::crypto::SignatureDigestAlgorithm;
 use sigstore::tuf::SigstoreRepository;
 use std::boxed::Box;
@@ -93,7 +92,7 @@ struct Cli {
     image: String,
 }
 
-async fn run_app() -> anyhow::Result<Vec<SignatureLayer>> {
+async fn run_app() -> anyhow::Result<VerificationConstraintVec> {
     let cli = Cli::parse();
 
     // Note well: this a limitation deliberately introduced by this example.
@@ -218,23 +217,23 @@ async fn run_app() -> anyhow::Result<Vec<SignatureLayer>> {
         .trusted_signature_layers(auth, &source_image_digest, &cosign_signature_image)
         .await?;
 
-    sigstore::cosign::filter_signature_layers(&trusted_layers, verification_constraint)
+    sigstore::cosign::filter_constraints(&trusted_layers, verification_constraint)
         .map_err(|e| anyhow!("{}", e))
 }
 
 #[tokio::main]
 pub async fn main() {
-    let trusted_signatures: anyhow::Result<Vec<SignatureLayer>> = run_app().await;
+    let unsatisfied_constraints: anyhow::Result<VerificationConstraintVec> = run_app().await;
 
-    std::process::exit(match trusted_signatures {
-        Ok(signatures) => {
-            if signatures.is_empty() {
-                eprintln!("Image verification failed: no matching signature found.");
-                1
-            } else {
+    std::process::exit(match unsatisfied_constraints {
+        Ok(unsatisfied_constraints) => {
+            if unsatisfied_constraints.is_empty() {
                 println!("Image successfully verified");
-                serde_json::to_writer_pretty(std::io::stdout(), &signatures).unwrap();
                 0
+            } else {
+                eprintln!("Image verification failed: not all constraints satisfied.");
+                // serde_json::to_writer_pretty(std::io::stdout(), &unsatisfied_constraints).unwrap();
+                1
             }
         }
         Err(err) => {
--- a/examples/verify/main.rs
+++ b/examples/verify/main.rs
@@ -16,10 +16,11 @@
 extern crate sigstore;
 use sigstore::cosign::verification_constraint::{
     AnnotationVerifier, CertSubjectEmailVerifier, CertSubjectUrlVerifier, PublicKeyVerifier,
-    VerificationConstraintVec,
+    VerificationConstraint, VerificationConstraintVec,
 };
-use sigstore::cosign::CosignCapabilities;
+use sigstore::cosign::{CosignCapabilities, SignatureLayer};
 use sigstore::crypto::SignatureDigestAlgorithm;
+use sigstore::errors::SigstoreVerifyConstraintsError;
 use sigstore::tuf::SigstoreRepository;
 use std::boxed::Box;
 use std::convert::TryFrom;
@@ -92,7 +93,7 @@ struct Cli {
     image: String,
 }
 
-async fn run_app() -> anyhow::Result<VerificationConstraintVec> {
+async fn run_app() -> anyhow::Result<(Vec<SignatureLayer>, Vec<Box<dyn VerificationConstraint>>)> {
     let cli = Cli::parse();
 
     // Note well: this a limitation deliberately introduced by this example.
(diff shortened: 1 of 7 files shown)


examples/verify/main.rs
```diff
@@ -218,23 +217,23 @@ async fn run_app() -> anyhow::Result<Vec<SignatureLayer>> {
...
-            } else {
+    std::process::exit(match unsatisfied_constraints {
+        Ok(unsatisfied_constraints) => {
+            if unsatisfied_constraints.is_empty() {
                 println!("Image successfully verified");
-                serde_json::to_writer_pretty(std::io::stdout(), &signatures).unwrap();
                 0
+            } else {
+                eprintln!("Image verification failed: not all constraints satisfied.");
+                println!("{:?}", unsatisfied_constraints);
```
viccuad, line 235: See TODO.
I'm undecided on how to continue here. Making `VerificationConstraint` Serializable seems doable (apart from working with boxed dyns..). The only hiccup is `ring::signature::VerificationAlgorithm` used in `PublicKeyVerifier`.

src/cosign/mod.rs
```diff
@@ -86,41 +86,47 @@ pub trait CosignCapabilities {
...
-                        }
-                    })
-                };
-                is_a_match
-            })
-            .cloned()
-            .collect();
-
-    if layers.is_empty() {
+) -> Result<VerificationConstraintVec> {
```
viccuad, line 101: I decided to not return the `SignatureLayers` that passed verification (or failed, neither), as I didn't see much value in having them, without having the associated list of satisfied or unsatisfied constraints to them.

That said, it would be easy to just construct a couple of vectors with the Layers in the already existing iterations, and it wouldn't change the complexity in time.

If wanted, we could return instead a 2d matrix that maps the relationship between layers and constraints (if a constraint has no associated layers there, is unsatisfied. If any constraint is unsatisfied, the verification fails).

examples/verify/main.rs
viccuad: Went with `Result<(),SigstoreFilterConstraintsError>`, see https://github.com/sigstore/sigstore-rs/pull/42/commits/34db8715bf0023bad652a60add2b8c4192e0db1a.

src/cosign/mod.rs
viccuad: Went with `Result<(),SigstoreFilterConstraintsError>`, see https://github.com/sigstore/sigstore-rs/pull/42/commits/34db8715bf0023bad652a60add2b8c4192e0db1a.

src/cosign/mod.rs
```diff
@@ -86,41 +86,53 @@ pub trait CosignCapabilities {
...
+/// aren't satisfied by the layers.
+///
+/// If there's any unsatisfied constraints it means that the image failed
+/// verification.
+/// If there's no unsatisfied constraints it means that the image passed
+/// verification.
+///
+/// Returns a Result with either Ok() for passed verification or
+/// SigstoreError::SigstoreFilterConstraintsError{}, which contains a vec of
+/// references to unsatisfied constraints.
```
flavio, line 99:
```suggestion
/// Returns a `Result` with either `Ok()` for passed verification or
/// [`SigstoreFilterConstraintsError`](crate::errors::SigstoreFilterConstraintsError), which contains a vector of
/// references to unsatisfied constraints.
```

I improved the formatting, the link should work too. Double check by running `cargo doc --open`

src/cosign/mod.rs
```diff
@@ -86,41 +86,53 @@ pub trait CosignCapabilities {
...
-    if layers.is_empty() {
-        Err(SigstoreError::SigstoreNoVerifiedLayer)
+pub fn filter_constraints<'a, 'b, I>(
+    signature_layers: &'a [SignatureLayer],
+    constraints: I,
+) -> std::result::Result<(), SigstoreFilterConstraintsError<'b>>
+where
+    I: Iterator<Item = &'b Box<dyn VerificationConstraint>>,
+{
+    let unsatisfied_constraints: Vec<&Box<dyn VerificationConstraint>> = constraints.filter(|c| {
```
flavio, line 110: Here we can replace `Vec<&Box<dyn VerificationConstraint>>` with the custom type `VerificationConstraintRefVec` we defined

src/cosign/mod.rs
```diff
@@ -86,41 +86,53 @@ pub trait CosignCapabilities {
...
+                    warn!(error = ?e, constraint = ?c, "Skipping layer because constraint verification returned an error");
+                    // handle errors as verification failures
+                    is_c_unsatisfied = true;
+                    false // keep searching to see if other layer satisfies
+                }
+            }
+        });
+        is_c_unsatisfied // if true, constraint gets filtered into result
+    }).collect();
+    if unsatisfied_constraints.is_empty() {
```
flavio, line 130: nitpicking, maybe add an empty line?

src/cosign/mod.rs
```diff
@@ -262,20 +272,67 @@ Hr/+CxFvaJWmpYqNkLDGRU+9orzh5hI2RrcuaQ==
...
+            email: wrong_email.clone(),
+            issuer: Some(issuer), // correct issuer
         };
         constraints.push(Box::new(vc));
 
         let vc = CertSubjectEmailVerifier {
-            email: email_constraint.clone(),
-            issuer: None,
+            email: wrong_email,
+            issuer: None, // missing issuer
```
flavio, line 282:
```suggestion
            issuer: None, // missing issuer - makes check more relaxed
```

The missing issuer per se, is not an error. This makes the verifier more relaxed. I've added the little clarification to ensure other readers don't assume this could lead to a verification == false

src/cosign/mod.rs
```diff
@@ -86,41 +86,53 @@ pub trait CosignCapabilities {
...
-                    })
-                };
-                is_a_match
-            })
-            .cloned()
-            .collect();
-
-    if layers.is_empty() {
-        Err(SigstoreError::SigstoreNoVerifiedLayer)
+pub fn filter_constraints<'a, 'b, I>(
```
flavio, line 103: As discussed via chat, we both agreed to go on with `verify_constraints`, which might return a `SigstoreVerifyConstraintsError`

flavio (changes requested):
LGTM, thanks for the hard work. I left some small comments to be addressed

src/cosign/mod.rs
viccuad: Done via force-push, since commit contents needed to change too.

examples/verify/main.rs
```diff
@@ -93,7 +93,7 @@ struct Cli {
     image: String,
 }
 
-async fn run_app() -> anyhow::Result<Vec<SignatureLayer>> {
+async fn run_app() -> anyhow::Result<(Vec<SignatureLayer>, Vec<Box<dyn VerificationConstraint>>)> {
```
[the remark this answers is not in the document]
viccuad: True! done in https://github.com/sigstore/sigstore-rs/pull/42/commits/9f56fd64da567c8b0ebabf9d090717887e38f86f.

src/cosign/mod.rs
```diff
@@ -52,9 +52,9 @@ pub use self::client::Client;
 
 pub mod client_builder;
 pub use self::client_builder::ClientBuilder;
+use self::verification_constraint::VerificationConstraint;
```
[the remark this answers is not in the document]
viccuad: True! done in https://github.com/sigstore/sigstore-rs/pull/42/commits/9f56fd64da567c8b0ebabf9d090717887e38f86f.

src/cosign/mod.rs
viccuad: Oh, good point! Fixed that one (and another one), and added `cargo doc` as part of `make test` in https://github.com/sigstore/sigstore-rs/pull/42/commits/167114e21b7590569d878ed3126e4349ea5ea01b.

src/cosign/mod.rs
viccuad: True! done in https://github.com/sigstore/sigstore-rs/pull/42/commits/9f56fd64da567c8b0ebabf9d090717887e38f86f.

src/cosign/mod.rs
viccuad: True! done in https://github.com/sigstore/sigstore-rs/pull/42/commits/9f56fd64da567c8b0ebabf9d090717887e38f86f.

src/cosign/mod.rs
viccuad: True! done in https://github.com/sigstore/sigstore-rs/pull/42/commits/9f56fd64da567c8b0ebabf9d090717887e38f86f.

flavio (approved):
LGTM. Thanks for the fixes

src/lib.rs
```diff
@@ -124,26 +125,26 @@
...
+//!   match result {
+//!       Ok(()) => {
+//!           println!("Image successfully verified");
+//!       }
+//!       Err(SigstoreVerifyConstraintsError {
+//!           unsatisfied_constraints,
+//!       }) => {
+//!           if unsatisfied_constraints.is_empty() {
+//!               println!("Image successfully verified");
+//!           } else {
```
flavio, line 143: In theory this should never happen. How could we end up having an empty list of unsatisfied constraints? I would remove this code block by dropping the `if/else` and use just what is currently into the `else` block

src/lib.rs
```diff
@@ -124,26 +125,26 @@
 //!     Box::new(pub_key_verifier),
 //!   ];
 //!
-//!   // Filter all the trusted layers, find the ones satisfying the constraints
-//!   // we just defined
-//!   let signatures_matching_requirements = filter_signature_layers(
+//!   // Filter the constraints, find the ones not satisfied by the layers on
+//!   // the image
```
flavio, line 129:
```suggestion
Use the given list of constraints to verify the trusted `signature_layers`. This will raise an error if one or more verification constraints are not satisfied.

flavio (changes requested):
Last comment, I promise :)

src/lib.rs
viccuad: true, good point. Fixed in https://github.com/sigstore/sigstore-rs/pull/42/commits/261c904a406d2b98ad021ee733ecded84dd32cbc.

src/lib.rs
viccuad: Fixed in https://github.com/sigstore/sigstore-rs/pull/42/commits/261c904a406d2b98ad021ee733ecded84dd32cbc.

flavio (approved):
Everything looks good, thanks!

Issue #32, opened 2022-02-07
Use right fulcio certificate when verifying bundles


flavio:
**Description**

Fulcio has updated its certificate. Right now, inside of Sigstore's TUF repository there are 2 certificates:

* `fulcio.crt.pem`
* `fulcio_v1.crt.pem`

Keyless signatures produced some months ago have been signed with `fulcio.crt.pem`, while more recent ones have been signed with `fulcio_v1.crt.pem`.

Right now the `cosign::Client` uses just one certificate, which causes some signatures to fail verification.

Apparently the `cosign` client isn't affected by this issue. We should understand how the upstream client handles that, and replicate the behaviour here.

flavio:
I found the root cause, to solve that we need to verify the certificate against a set of root CAs.

After a lot of searching, I found out that `webpki` can help us there. Unfortunately it needs some minor changes that are implemented via https://github.com/briansmith/webpki/pull/254

asraa:
> I found the root cause, to solve that we need to verify the certificate against a set of root CAs.

Yep!

> Apparently the cosign client isn't affected by this issue. We should understand how the upstream client handles that, and replicate the behaviour here.

Also -- Hayden and I scoped how this would look like in the future as more Fulcio and rekor logs are rotated. It will end up being that we add all targets that have an annotation `"usage": Fulcio" attached to them. Hayden made a PR here  https://github.com/sigstore/cosign/pull/1423

I wonder if awslabs has support for parsing/reading custom metadata attached to targets? I can try to look into that

flavio:
> Also -- Hayden and I scoped how this would look like in the future as more Fulcio and rekor logs are rotated. It will end up being that we add all targets that have an annotation `"usage": Fulcio" attached to them. Hayden made a PR here [sigstore/cosign#1423](https://github.com/sigstore/cosign/pull/1423)

Thanks for pointing this out! In the meantime my plan is to just load all the files that are matching a certain pattern.
We can change the code once the TUF metadata is updated. Can you ping me once this is out?
 
> I wonder if awslabs has support for parsing/reading custom metadata attached to targets? I can try to look into that

Yes, it seems this is handled: https://docs.rs/tough/latest/tough/schema/struct.Target.html - I guess they will show up inside of the `custom` attribute, if not, they will be inside of the `_extra` one, which is populated by "serde flatten".

There's another question I have. What is your plan regarding rekor's public key? What do you plan to do if this key is suddenly changed?

[...]

PR #43, opened 2022-03-07
fix handling of fulcio certificates

flavio:
#### Summary

Fulcio has recently introduced a new root CA to issue certificates. Prior to this commit, the code was making use only of the initial root CA.
Because of that, recent keyless signatures were considered not valid.

Now the code is taking into account of the root CAs that Fulcio used over the time.

#### Ticket Link

Fixes https://github.com/sigstore/sigstore-rs/issues/32

#### Release Note

```release-note
* Fix(TUF): fetch all the Fulcio certificates found inside of the TUF repository
* Fix(Verification): ensure all the Fulcio root certificates are used to verify signatures. This fixes issue #32
```

### DIFF
--- a/src/cosign/client.rs
+++ b/src/cosign/client.rs
@@ -14,7 +14,6 @@
 // limitations under the License.
 
 use async_trait::async_trait;
-use x509_parser::{traits::FromDer, x509::SubjectPublicKeyInfo};
 
 use super::{
     constants::SIGSTORE_OCI_MEDIA_TYPE,
@@ -22,8 +21,11 @@ use super::{
     CosignCapabilities,
 };
 use crate::crypto::CosignVerificationKey;
-use crate::errors::{Result, SigstoreError};
 use crate::registry::Auth;
+use crate::{
+    crypto::certificate_pool::CertificatePool,
+    errors::{Result, SigstoreError},
+};
 
 /// Cosign Client
 ///
@@ -31,7 +33,7 @@ use crate::registry::Auth;
 pub struct Client {
     pub(crate) registry_client: Box<dyn crate::registry::ClientCapabilities>,
     pub(crate) rekor_pub_key: Option<CosignVerificationKey>,
-    pub(crate) fulcio_pub_key_der: Option<Vec<u8>>,
+    pub(crate) fulcio_cert_pool: Option<CertificatePool>,
 }
 
 #[async_trait]
@@ -81,20 +83,12 @@ impl CosignCapabilities for Client {
             }
         };
 
-        let fulcio_pub_key = match &self.fulcio_pub_key_der {
-            None => None,
-            Some(der) => {
-                let (_, key) = SubjectPublicKeyInfo::from_der(der)?;
-                Some(key)
-            }
-        };
-
         build_signature_layers(
             &image_manifest,
             source_image_digest,
             &layers,
             self.rekor_pub_key.as_ref(),
-            fulcio_pub_key.as_ref(),
+            self.fulcio_cert_pool.as_ref(),
         )
     }
 }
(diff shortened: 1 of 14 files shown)


src/cosign/client.rs
```diff
@@ -53,7 +55,7 @@ impl CosignCapabilities for Client {
             "{}/{}:{}.sig",
             image_reference.registry(),
             image_reference.repository(),
-            manifest_digest.replace(":", "-")
+            manifest_digest.replace(':', "-")
```
ereslibre, line 58: Shouldn't both be quoted with `'`?

src/cosign/client_builder.rs
```diff
@@ -29,47 +29,66 @@ use crate::registry::ClientConfig;
...
         self.rekor_pub_key = Some(key.to_string());
         self
     }
 
-    /// Specify the certificate used by Fulcio.
+    /// Specify the certificate used by Fulcio. This method can be invoked
+    /// multiple times to add all the certificates that Fulcio used over the
+    /// time.
+    ///
+    /// `cert` is a PEM encoded cerificate
```
ereslibre, line 67:
```suggestion
    /// `cert` is a PEM encoded certificate
```

src/cosign/client_builder.rs
```diff
@@ -29,47 +29,66 @@ use crate::registry::ClientConfig;
...
+
+    /// Specify the certificates used by Fulcio. This method can be invoked
+    /// multiple times to add all the certificates that Fulcio used over the
+    /// time.
+    ///
+    /// The certificates can be obtained by using the helper methods under the
+    /// [`tuf`](crate::tuf) module.
+    ///
+    /// When provided, this enables Fulcio's integration.
+    pub fn with_fulcio_certs(mut self, certs: &[crate::registry::Certificate]) -> Self {
```
ereslibre, line 90: What about `add_fulcio_certs` so it's clear from the API perspective (aside from the documentation) that this is an additive operation? `with_fulcio_certs`, and without reading the docs, it might seem that this operation is replacing previous added certs.

src/tuf/repository_helper.rs
```diff
@@ -50,23 +50,46 @@ impl RepositoryHelper {
         })
     }
 
-    /// Fetch Fulcio certificate from the given TUF repository or reuse
+    /// Fetch Fulcio certificates from the given TUF repository or reuse
     /// the local cache is used if its contents are not outdated.
```
ereslibre, line 54:
```suggestion
    /// the local cache if its contents are not outdated.
```

ereslibre (approved):
Thanks @flavio! Some minor comments.

src/cosign/client_builder.rs
flavio: You're right. I'll change the behaviour of `with_fulcio_certs` to replace the previously added ones.

src/cosign/client.rs
flavio: This is counterintuitive, [str::replace](https://doc.rust-lang.org/std/primitive.str.html#method.replace) has a generic first parameter, while the second one is a `&str`.

As a result of that, clippy will complain when the first param is a "useless" `str` (like `":"`), instead of being a simple char

viccuad (approved):
LGTM too modulo Rafa's comments.

src/cosign/client_builder.rs
flavio: solved

flavio:
thanks for the reviews, I addressed all the comments.

@lukehinds: could you take a look please?

lukehinds (approved):
looks fine from my side.

flavio:
Thanks, I've fixed the conflicts with the `main` branch

PR #44, opened 2022-03-08
Ensure docs are automatically published

flavio:
The `docs` GitHub Action job has been failing lately. We have changed the default settings of the `GITHUB` token to have read-only capabilities.

This breaks the `docs` action, because the action performs the following operation:

  * Checkout the code
  * Render the docs
  * Push the new docs to a dedicated branch of the repository

The last step is broken by the lack of write permissions associated with the GITHUB token.

This commits adds the permissions that are required to perform commits.

Hopefully this is going to be enough, if not I'll create a new PR which instead uses a deployment key.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+permissions:
+  deployments: write
+
 jobs:
   deploy:
     runs-on: ubuntu-latest


lukehinds:
I know there was an org wide plan to implement `permissions: deployments: read`, not sure if we can overwrite in an individual repo basis. This was to protect secrets from leaking.  

cc @cpanato / @bobcallaway

lukehinds (approved):

flavio:
Unfortunately it didn't work: https://github.com/sigstore/sigstore-rs/runs/5468448546?check_suite_focus=true

I'll revert the commit and create a new PR that instead uses a deployment key. That means I'll have to add a deployment key to the repository. I guess there's no problem with that... 

CC @lukehinds

PR #50, opened 2022-04-08
OCI caching client

flavio:
Allow end consumers of the library to create a `cosign::Client` that caches the response it gets from OCI registries.

The cache entries are automatically expired after 60 seconds.

This can be useful in certain scenarios, like when writing code that can verify the same container image, but with different constraints.

Imagine the following scenario: write a REST API that exposes sigstore verification capabilities.
In this case, the REST API can receive multiple verification requests of the same image, but each time with slightly different verification constraints.
Thanks to the registry caching feature, the REST server would significantly reduce the number of interactions with the remote OCI registry. That would lead to shorter response times and higher request per seconds.

To be completely honest, this is something we need inside of kubewarden, but I think others could benefit from that. If this is not something relevant for this library we could just move this code into kubewarden :peace_symbol: 

## Demo

Running without caching:

```console
╰─ cargo run --example verify -- --use-sigstore-tuf-data --loops 3 ghcr.io/kubewarden/policies/allowed-fsgroups-psp:v0.1.3
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/examples/verify --use-sigstore-tuf-data --loops 3 'ghcr.io/kubewarden/policies/allowed-fsgroups-psp:v0.1.3'`
2022-04-08T12:30:14.688800Z  INFO verify: Downloading data from Sigstore TUF repository
Loop 1/3
Image successfully verified
Elapsed: 1.84s
------
Loop 2/3
Image successfully verified
Elapsed: 1.38s
------
Loop 3/3
Image successfully verified
Elapsed: 1.58s
------
```

Using caching:

```console
╰─ cargo run --example verify -- --use-sigstore-tuf-data --loops 3 --enable-registry-caching ghcr.io/kubewarden/policies/allowed-fsgroups-psp:v0.1.3
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/examples/verify --use-sigstore-tuf-data --loops 3 --enable-registry-caching 'ghcr.io/kubewarden/policies/allowed-fsgroups-psp:v0.1.3'`
2022-04-08T12:30:59.931631Z  INFO verify: Downloading data from Sigstore TUF repository
Loop 1/3
Image successfully verified
Elapsed: 1.55s
------
Loop 2/3
Image successfully verified
Elapsed: 21.72ms
------
Loop 3/3
Image successfully verified
Elapsed: 16.23ms
------
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ rustls-tls = ["oci-distribution/rustls-tls", "openidconnect/rustls-tls"]
 [dependencies]
 async-trait = "0.1.52"
 base64 = "0.13.0"
+cached = "0.34.0"
 lazy_static = "1.4.0"
 # TODO: go back to the officially release oci-distribution once these patches are released
 oci-distribution = { git = "https://github.com/krustlet/oci-distribution", rev = "0f717968093a5415f428503d741dedf24ea97948", default-features = false }
--- a/examples/verify/main.rs
+++ b/examples/verify/main.rs
@@ -24,6 +24,7 @@ use sigstore::errors::SigstoreVerifyConstraintsError;
 use sigstore::tuf::SigstoreRepository;
 use std::boxed::Box;
 use std::convert::TryFrom;
+use std::time::Instant;
 
 extern crate anyhow;
 use anyhow::anyhow;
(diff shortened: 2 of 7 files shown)


viccuad (approved):
LGTM! This is a nice addition indeed, thanks.

PR #52, opened 2022-04-22
No need to unwrap as we return within a Result

lukehinds:
Remove unwrap or else and instead map_err and bubble up with the ?
operator.

#### Ticket Link

Fixes

#### Release Note

```release-note

```

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -144,4 +144,7 @@ pub enum SigstoreError {
 
     #[error("Verification of OIDC claims received from OpenIdProvider failed")]
     ClaimsVerificationError,
+
+    #[error("Failed to access token endpoint")]
+    ClaimsAccessPointError,
 }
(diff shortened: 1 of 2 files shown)


viccuad (approved):
LGTM, but haven't touched this part :). Could be a good moment to also create an error for `Failed to discover OpenID Provider` in `OpenIDAuthorize.auth_url()`, and we are done with both `OpenIDAuthorize` and `RedirectListener`, besides broad panics.

lukehinds:
> LGTM, but haven't touched this part :). Could be a good moment to also create an error for `Failed to discover OpenID Provider` in `OpenIDAuthorize.auth_url()`, and we are done with both `OpenIDAuthorize` and `RedirectListener`, besides broad panics.

I will do a follow up PR there as I plan to refactor that could some more

PR #53, opened 2022-04-22
Remove last unwrap or else and only use SigstoreError

lukehinds:


viccuad (approved):
thanks :). I would prefer the commits squashed into 2 or 2, as the follow-up commits are trivial fixes.

lukehinds:
> thanks :). I would prefer the commits squashed into 2 or 2, as the follow-up commits are trivial fixes.

I use the github squash and merge feature.

PR #54, opened 2022-04-27
Return the ID token

lukehinds:
Alongside the claims (containing the scopes) we also need the`id_token`.

Also changes `println!` with `tracing::error`

### DIFF
--- a/src/oauth/openidflow.rs
+++ b/src/oauth/openidflow.rs
@@ -250,7 +250,8 @@ impl RedirectListener {
 
                 let id_token = token_response
                     .extra_fields()
-                    .id_token().ok_or(SigstoreError::NoIDToken)?;
+                    .id_token()
+                    .ok_or(SigstoreError::NoIDToken)?;
 
                 let id_token_verifier: CoreIdTokenVerifier = self.client.id_token_verifier();
 
--- a/src/oauth/openidflow.rs
+++ b/src/oauth/openidflow.rs
@@ -84,7 +84,8 @@ use crate::errors::{Result, SigstoreError};
 use tracing::error;
 
 use openidconnect::core::{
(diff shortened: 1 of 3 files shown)


src/oauth/openidflow.rs
```diff
@@ -246,17 +247,24 @@ impl RedirectListener {
                     .request(http_client)
                     .map_err(|_| SigstoreError::ClaimsAccessPointError)?;
 
+                let id_token = token_response
+                    .extra_fields()
+                    .id_token()
+                    .expect("Failed to get id_token")
```
flavio, line 253: Maybe it would be better to introduce a custom Error type and just return it, instead of causing a panic. That gives the end consumer of the library the ability to handle errors the way he wants

flavio (dismissed):
LGTM, I just left a minor suggestion

src/oauth/openidflow.rs
lukehinds: I will have a play with this, its an Option, not a Result, so there is no err (only None) to create a map_err. Any ideas?

flavio: You can do:

```rust
let foo = token_response.extra_fields().id_token().ok_or_else(|| SigstoreError::NoIDToken())?;
```

lukehinds: thanks @flavio , implemented in 09abdd36fd7559715ec21c8a9a3198813d4988a9

flavio (commented):
Cool, just fix the clippy warning and the merge it :)

lukehinds:
> Fix clippy error on ok_or_else

woops, i need to set up some sort of pre-commit hook. fixed in 9d10b1a587b6e58e5bf973f372d2bfaf7c6abdc7

flavio (approved):

PR #58, opened 2022-05-10
pin the github actions to use the git hash

cpanato:
#### Summary
as a best practice, we should pin the actions to use the git hash instead of the tag or `main`.
also, this makes it similar to the other workflows we have in the sigstore org.

- pin the github actions to use the git hash

#### Ticket Link
n/a

#### Release Note

```release-note
NONE
```

### DIFF
--- a/.github/workflows/auto-publish-crates-upon-release.yml
+++ b/.github/workflows/auto-publish-crates-upon-release.yml
@@ -7,8 +7,8 @@ jobs:
   publish-automatically:
     runs-on: ubuntu-latest
     steps:
-      -  uses: actions/checkout@v2
-      -  uses: actions-rs/toolchain@v1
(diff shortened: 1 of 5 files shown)


cpanato:
this PR superseded this https://github.com/sigstore/sigstore-rs/pull/56

flavio (approved):
LGTM. 

@cpanato do you know if dependabot can propose updates even when the action versions are referenced by commit id?

cpanato:
> LGTM.
> 
> @cpanato do you know if dependabot can propose updates even when the action versions are referenced by commit id?

it can :) the only drawback for now is that does not update the version comment, but we do that manually for now in most of the cases. there is an open issue for dependabot update the comment as well, so i guess it is just a matter of time :)

PR #60, opened 2022-05-11
Implements CI check for unsafe, unwrap and panic

lukehinds:
Performs a search of the `src` directory (not `examples` etc) for
the terms that can be problematic in production code.

If a term is used legitimately, the use can be skipped:

`something.unwrap(); //#[allow_ci]`

I have currently commented out all instances in test code or where
a comment is provided to explain why use of function is acceptable

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,8 +32,8 @@ jobs:
           command: test
           args: --workspace
 
-  fmt:
-    name: Rustfmt
+  static:
+    name: Static code checks
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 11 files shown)


tarcieri:
Note that built-in lints + clippy can do this for you:

https://rust-lang.github.io/rust-clippy/master/#unwrap_used
https://rust-lang.github.io/rust-clippy/master/#panic
https://rust-lang.github.io/rust-clippy/master/#panic_in_result_fn

I'd recommend `#[forbid(unsafe_code)]` to prevent unsafe from being used at a crate-level

lukehinds:
@tarcieri that' very useful thanks! 

Let me play with this. One of things I have found with clippy is you can only decorate a whole function and not a single line, e.g.

```
#[allow(clippy::unwrap_used)]
fn main() {
    let _ = std::fs::read_dir(".").unwrap();
}
```

lukehinds:
actually this is a pretty decent, it automatically ignores tests (which was my concern). I will close this PR and lets go native with clippy

PR #71, opened 2022-06-23
Add Fulcio intermediate CA certificate to intermediate pool

### LINKED ISSUE #70: Verification error: Not issued by a  trusted root error
**Description**

Keyless verification is not working since the Fulcio Intermediate CA certificate was introduced, see https://github.com/sigstore/cosign/pull/1774

This is the error I see:
```
INFO sigstore::cosign::signature_layers: Skipping OCI layer because of error error=CertificateValidityError("Not issued by a  trusted root")           
```

If the signature is stored in an OCI, the chain is stored in the annotation `dev.sigstore.cosign/chain`

**Version**

sigstore-rs v0.3.1

raulcabello:
#### Summary

Add an Intermediate CA certificate that is needed for building a chain from a leaf certificate. Verification is currently not working because Fulcio is issuing certificates with this intermediate certificate.

For the V3 TUF Root, this intermediate certificate will be bundled, so that it is easily discoverable and revokable. For now, we'll simply bundle it in the code.

Fixes #70

### DIFF
--- a/src/crypto/certificate_pool.rs
+++ b/src/crypto/certificate_pool.rs
@@ -18,16 +18,35 @@ use crate::{
     registry::Certificate,
 };
 
+// The untrusted intermediate CA certificate, used for chain building
+// TODO: Remove once this is bundled in TUF metadata.
+const FULCIO_INTERMEDIATE_V1: &str = "-----BEGIN CERTIFICATE-----
+MIICGjCCAaGgAwIBAgIUALnViVfnU0brJasmRkHrn/UnfaQwCgYIKoZIzj0EAwMw
+KjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTAeFw0y
+MjA0MTMyMDA2MTVaFw0zMTEwMDUxMzU2NThaMDcxFTATBgNVBAoTDHNpZ3N0b3Jl
+LmRldjEeMBwGA1UEAxMVc2lnc3RvcmUtaW50ZXJtZWRpYXRlMHYwEAYHKoZIzj0C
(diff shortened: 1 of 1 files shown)


src/crypto/certificate_pool.rs
```diff
@@ -83,4 +118,20 @@ impl CertificatePool {
...
+
+    fn create_chains_for_all_certificates(&self) -> Vec<Vec<&picky::x509::Cert>> {
+        let mut chains: Vec<Vec<&picky::x509::Cert>> = vec![];
+        self.trusted_roots.iter().for_each(|trusted_root| {
+            chains.push([trusted_root].to_vec());
+        });
+        self.intermediates.iter().for_each(|intermediate| {
+            for root in self.trusted_roots.iter() {
+                if root.is_parent_of(intermediate).is_ok() {
+                    chains.push([intermediate, root].to_vec());
```
viccuad, line 130: shouldn't we be pushing only the intermediate certificate? After all, the root should be already in the chain.

flavio: A chain must start with the intermediate and then add all the certificates that lead to the root one

src/crypto/certificate_pool.rs
```diff
@@ -66,14 +99,16 @@ impl CertificatePool {
...
-                .exact_date(&cert.valid_not_before())
-                .verify()
-                .is_ok()
-        });
+        let verified = self
+            .create_chains_for_all_certificates()
+            .iter()
+            .any(|chain| {
+                cert.verifier()
+                    .chain(chain.iter().cloned())
```
flavio, line 107: `chain` is a `Vec<&Cert>`, when we call `iter` on it we end up iterating over `&&cert`.

Instead of cloning them, we could do something like:

```rust
.chain(chain.iter().map(|c| *c))
```

which is equivalent to:

```suggestion
                    .chain(chain.iter().copied())
```

I think this works and should be faster than clone. Can you double check please?

PR #72, opened 2022-06-23
Tag v0.3.2

flavio:
Changes:

  * Add Fulcio intermediate CA certificate to intermediate pool - fixes issue #70


flavio:
> LGTM, but we may like to bump Cargo.lock too, instead of letting dependabot do it.

`Cargo.lock` is not under version control because this crate is a library. Hence there's no work to be done :+1:

Issue #27, opened 2022-02-03
Key generation and signer interface


lukehinds:
Much like the go counterpart in sigstore/sigstore, implement key generation and a signer interface. This should provide the following:

1. Generate of a key pair using a provided alg (rsa, ecdsa)
2. public key derive from private key
3. Signing interface
4. Serialisation (handle PEM / DER en/decode etc).

lukehinds:
Will keep this open to track Ed25519 and RSA modules

- [ ] ecdsa
- [ ] Ed25519
- [ ] RSA

Many thank @Xynnn007 for taking this on!

Xynnn007:
@lukehinds Hi, ed25519 has been supported in https://github.com/sigstore/sigstore-rs/pull/87/commits/8314f8f7867e0e80c2e2ca0470a6897f859ae93c

lukehinds:
Just going to add one more @Xynnn007 , some examples (for users) would be very useful. See examples in the root folder.

Xynnn007:
Also, now `CosignVerificationKey` has the similar semantics with the `SigStoreSigner`'s public key. I'd prefer to add the following for ease
* `Into` trait for `SigStoreSigner` or `KeyPair`
* `From` trait for `CosignVerificationKey`

lukehinds:
sounds good @Xynnn007

Xynnn007:
There might be some refactoring for the key_interface, to deal with the following problems:
* Now when users use the `signing_key` module, they must import the crates into the `Cargo.toml`, like `p256`, `p384`, etc. I'd prefer users do not need to care about the underlying Elliptic Curves. They can directly use `sigstore` crate without import any other crates to do all the key related operations, thus we need to do some wrapping. 
* Now we do not have a convenient call routine for a user to import a private key and convert it into a `SigStoreSigner`. Also `SigStoreSigner` is somehow overlapped functionally with `Signer` trait object. I'd like to combine them, and give a more easy way for users to convert a PEM/DER encoded private key into an object who can sign.

lkatalin:
@Xynnn007 Strongly agree about not having users import extra crates, thank you!

Xynnn007:
Yea, in a sense `CosignVerificationKey` struct has the same issue that when use this we need to import `ring` crate. How about refactor it into an enum, the enum values will be the same as `crypto::signing_key::SigningScheme`, s.t.
```rust
pub enum SigningScheme {
    // TODO: RSA schemes
[...]

PR #87, opened 2022-07-29
feat: add signing key module

Xynnn007:
#### Summary

Due to https://github.com/sigstore/sigstore-rs/issues/27, `sigstore-rs` needs key generation and signing interfaces.

This pr aims to give a framework for signer and key pairs.

The KDF refers to [go counterpart in sigstore/sigstore](https://github.com/sigstore/sigstore)

#### Release Note

- Definition of interface for key generation, derivation of private and public keys and signing.

The interface for `KeyPair` included:
```rust
fn public_key_to_pem(&self) -> Result<String>;
fn public_key_to_der(&self) -> Result<Vec<u8>>;
fn private_key_to_encrypted_pem(&self, password: &[u8]) -> Result<Zeroizing<String>>;
fn private_key_to_pem(&self) -> Result<Zeroizing<String>>;
fn private_key_to_der(&self) -> Result<Zeroizing<Vec<u8>>>;
```

And a `Signer`
```rust
fn key_pair(&self) -> &dyn KeyPair;
fn sign(&self, msg: &[u8]) -> Result<Vec<u8>>;
```

- ECDSA was implemented.
```rust
let ec_key_pair = EcdsaKeys::<p256::NistP256>::new().unwrap();
```

- For easy to use signing schemes, use an enum interface, like
```rust
let signer = SigStoreSigner::new(SigningScheme::ECDSA_P256_SHA256_ASN1).unwrap();
let sig = signer.sign(b"test message").unwrap();
```
#### Documentation

NONE


src/crypto/signing_key/kdf.rs
```diff
@@ -0,0 +1,253 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Key Derivation Function for Sigstore
+
+use rand::Rng;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use sodiumoxide::crypto::secretbox;
```
tarcieri, line 20: `sodiumoxide` is deprecated: https://github.com/sodiumoxide/sodiumoxide/commit/5bb1dfd2578539b89ffb0cbea25f21f00cfb963e

The `xsalsa20poly1305` crate provides the same construction as secretbox: https://github.com/RustCrypto/AEADs/tree/master/xsalsa20poly1305

Separately, Ii you're looking for a ready-made construction providing password-based encryption with scrypt-based key derivation, check out the `pkcs5` crate: https://github.com/RustCrypto/formats/tree/master/pkcs5

Xynnn007: Thank you very much for your advice! 

I'm not quite familiar with cryptography here. Here `sigstore` uses decrypt to derive key, and uses `xsalsa20poly1305` to encrypt & decrypt, which is not an example in [https://datatracker.ietf.org/doc/html/rfc2898#section-6.2](PKCS#5). Thus not supported in `pkcs5` crate, right?

lukehinds:
Thanks! A few pointers

* Some docs would be useful (https://sigstore.github.io/sigstore-rs/sigstore/) 
* We put example code for users to reference (https://github.com/sigstore/sigstore-rs/tree/main/examples) 

cc @flavio

src/crypto/signing_key/kdf.rs
tarcieri: Aah, ok, if this an existing scheme you can't use PKCS#5.

But the `xsalsa20poly1305` crate will work fine and eliminate the need for an unmaintained FFI binding to an external C library as a dependency.

Xynnn007: Thanks, I've changed to `xsalsa20poly1305`

Xynnn007:
> Thanks! A few pointers
> 
> * Some docs would be useful (https://sigstore.github.io/sigstore-rs/sigstore/)
> * We put example code for users to reference (https://github.com/sigstore/sigstore-rs/tree/main/examples)
> 
> cc @flavio

Thanks! I'd like to know whether this pr handles https://github.com/sigstore/sigstore-rs/issues/27?
If so, I'd like to carry on Ed25519 and RSA modules

Cargo.toml
```diff
@@ -36,7 +36,17 @@ tough = { version = "0.12.1", features = [ "http" ] }
 tracing = "0.1.31"
 url = "2.2.2"
 x509-parser = { version = "0.14.0", features = ["verify"] }
-
+zeroize = "*"
```
flavio: I think it would be better to require a specific version. That doesn't necessary mean to mention the patch level, but `major.minor` should be required.

Cargo.toml
```diff
@@ -36,7 +36,17 @@ tough = { version = "0.12.1", features = [ "http" ] }
...
+rand = "0.8.5"
+scrypt = "0.10.0"
+xsalsa20poly1305 = "*"
```
flavio: I think it would be better to require a specific version. That doesn't necessary mean to mention the patch level, but `major.minor` should be required.

Cargo.toml
```diff
@@ -36,7 +36,17 @@ tough = { version = "0.12.1", features = [ "http" ] }
...
+pkcs8 = { version = "0.9.0", features = ["pem", "alloc", "pkcs5", "encryption"] }
+elliptic-curve = { version = "*", features = [ "arithmetic", "pem" ] }
+p256 = "*"
+p384 = "*"
+ecdsa = { version = "*", features = [ "pkcs8", "digest" ] }
```
flavio: I think it would be better to require a specific version. That doesn't necessary mean to mention the patch level, but `major.minor` should be required.

src/crypto/signing_key/kdf.rs
```diff
@@ -0,0 +1,251 @@
...
+        Self {
+            name: NAME_SCRYPT.into(),
+            params: ScryptParams {
+                n: SCRYPT_N,
+                r: SCRYPT_R,
+                p: SCRYPT_P,
+            },
+            salt,
+        }
+    }
```
flavio, line 97: nit, it would be more idiomatic to make that the implementation of the `Default` trait

src/crypto/signing_key/kdf.rs
```diff
@@ -0,0 +1,251 @@
...
+pub const SCRYPT_P: u32 = 1;
+
+/// Secret box name
+pub const NAME_SECRET_BOX: &str = "nacl/secretbox";
+
+/// Key length for secretbox
+pub const BOX_KEY_SIZE: usize = 32;
+
+/// Nonce length for secretbox
+pub const BOX_NONCE_SIZE: u32 = 24;
```
flavio, line 46: nit, question: do all these constants need to be public? I did a quick search and some seem to be used only inside of this file

src/crypto/signing_key/kdf.rs
```diff
@@ -0,0 +1,251 @@
...
+
+impl SecretBoxCipher {
+    fn new() -> Self {
+        let nonce = generate_random(BOX_NONCE_SIZE);
+        Self {
+            name: NAME_SECRET_BOX.into(),
+            nonce,
+            encrypted: false,
+        }
+    }
```
flavio, line 139: nit, it would be more idiomatic to make that the implementation of the `Default` trait

src/crypto/signing_key/kdf.rs
```diff
@@ -0,0 +1,251 @@
...
+
+/// Key length for secretbox
+pub const BOX_KEY_SIZE: usize = 32;
+
+/// Nonce length for secretbox
+pub const BOX_NONCE_SIZE: u32 = 24;
+
+/// Parameters for scrypt algorithm.
+#[derive(Serialize, Deserialize)]
+pub struct ScryptParams {
```
flavio, line 50: Does it need to be `pub`?

src/crypto/signing_key/kdf.rs
```diff
@@ -0,0 +1,251 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::crypto::signing_key::kdf::Data;
+
+    /// This test will firstly deserialize the given KDF
+    /// payload generated from cosign in golang, and then
+    /// serialize the generated object into a new string.
+    #[test]
+    fn serde_kdf() {
+        let raw = r#"{"kdf":{"name":"scrypt","params":{"N":32768,"r":8,"p":1},"salt":"+QseLb/O/0j2dG201MALNSv2xLcclv6UvpXZVvXGT0k="},"cipher":{"name":"nacl/secretbox","nonce":"B5zH5d9AwoPkgaPAwIgpnft2BO6HZM/j"},"ciphertext":"RQPqIJtoWjlVC49xXNG+zfkGrJF3DWIhdRArI0XeTjGx04QzjAAeybGgW4T9JWKuYYe49NIZCEOD2G8cisMJ9KXHPaxT6Q/lLa8XrkavRrzkaD3xj8tc2AAntvUz8OACtH3zmimeFLr+EtecDb/UNjNFCtW1SlIh6DsfTsbBL67uQqLrFQMW8r70SvsZLkXV8mFhMsKyVryWlQ=="}"#;
```
flavio, line 246: nit, this could be improved, from a readability POV, by using the [`json!`](https://docs.rs/serde_json/latest/serde_json/macro.json.html) macro:

```rust
let data_json = json!({
  "kdf": {
    "name": "scrypt",
    "params": {
      "N": 32768,
      "r": 8,
      "p": 1
    },
    "salt": "+QseLb/O/0j2dG201MALNSv2xLcclv6UvpXZVvXGT0k="
  },
  "cipher": {
    "name": "nacl/secretbox",
    "nonce": "B5zH5d9AwoPkgaPAwIgpnft2BO6HZM/j"
  },
  "ciphertext": "RQPqIJtoWjlVC49xXNG+zfkGrJF3DWIhdRArI0XeTjGx04QzjAAeybGgW4T9JWKuYYe49NIZCEOD2G8cisMJ9KXHPaxT6Q/lLa8XrkavRrzkaD3xj8tc2AAntvUz8OACtH3zmimeFLr+EtecDb/UNjNFCtW1SlIh6DsfTsbBL67uQqLrFQMW8r70SvsZLkXV8mFhMsKyVryWlQ=="
});
```

Then, on the next line you can use the `serde_json::from_value` instead of the `serde_json::from_str`

src/crypto/signing_key/kdf.rs
```diff
@@ -0,0 +1,251 @@
...
+        let data: Data = serde_json::from_str(raw).unwrap();
+        let raw_new = serde_json::to_string(&data).unwrap();
+        assert_eq!(raw_new, raw);
```
flavio, line 249: Another nit... Looking better at this test, I would recommend to add a new developer dependency called [`assert-json-diff`](https://crates.io/crates/assert-json-diff) and then use the [`assert_json_eq!`](https://docs.rs/assert-json-diff/2.0.2/assert_json_diff/#exact-matching) macro.

The test code would be:

```rust
let input_json = json!(...) // the raw json you declared
let data: Data = serde_json::from_value(input_json).expect("Cannot deserialize json Data");
let actual_json = serde_json::to_value(&data).expect("Cannot serialize Data back to JSON");
assert_json_eq!(input_json, actual);
```

src/lib.rs
```diff
@@ -162,6 +162,7 @@
 
 #![forbid(unsafe_code)]
 #![warn(clippy::unwrap_used, clippy::panic)]
+#![allow(non_camel_case_types)]
```
flavio, line 165: Our of curiosity, why do we need that? nit, I would prefer that exception to be done on a per file basis, possibly close to the code that needs the exception

flavio (commented):
Awesome work :clap: 

I left some comments, mostly nits

Cargo.toml
Xynnn007: Yes, I have changed it in the most recent commit

flavio: Thanks!

src/crypto/signing_key/kdf.rs
Xynnn007: Yes, thanks!

src/crypto/signing_key/kdf.rs
Xynnn007: Good, thanks!

src/crypto/signing_key/kdf.rs
Xynnn007: Surely!

src/lib.rs
Xynnn007: Yes, I'll fix it

src/crypto/signing_key/kdf.rs
Xynnn007: That's such a cool dependency! thx!

src/crypto/signing_key/kdf.rs
Xynnn007: Sry, and I make the `kdf` public now

src/crypto/signing_key/ed25519.rs
```diff
@@ -0,0 +1,354 @@
...
+
+    /// Builds a `Ed25519Keys` from encrypted pkcs8 PEM-encided private key.
+    /// The label should be [`COSIGN_PRIVATE_KEY_PEM_LABEL`] or
+    /// [`SIGSTORE_PRIVATE_KEY_PEM_LABEL`].
+    pub fn from_encrypted_pem(encrypted_pem: &[u8], password: &[u8]) -> Result<Self> {
+        let key = pem::parse(encrypted_pem)?;
+        match &key.tag[..] {
+            COSIGN_PRIVATE_KEY_PEM_LABEL | SIGSTORE_PRIVATE_KEY_PEM_LABEL => {
+                let der = kdf::decrypt(&key.contents, password)?;
+                let pkcs8 = pkcs8::PrivateKeyInfo::try_from(&der[..]).map_err(|e| {
```
tarcieri, line 86: FYI, the `pkcs8` crate has built-in PEM support oriented at private keys via the `SecretDocument` type:

https://docs.rs/pkcs8/latest/pkcs8/struct.SecretDocument.html#method.from_pem

It uses [`pem-rfc7468`](https://docs.rs/pkcs8/latest/pkcs8/struct.SecretDocument.html#method.from_pem) which provides a constant-time PEM parser intended to mitigate [sidechannel attacks when parsing private keys](https://arxiv.org/pdf/2108.04600.pdf). It also handles zeroization on drop via the `zeroize` crate.

src/crypto/signing_key/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// `private_key_to_encrypted_pem` will export the encrypted asn.1 pkcs8 private key.
+    fn private_key_to_encrypted_pem(&self, password: &[u8]) -> Result<Zeroizing<String>>;
+
+    /// `private_key_to_pem` will export the PEM-encoded pkcs8 private key.
+    fn private_key_to_pem(&self) -> Result<Zeroizing<String>>;
+
+    /// `private_key_to_der` will export the asn.1 pkcs8 private key.
+    fn private_key_to_der(&self) -> Result<Zeroizing<Vec<u8>>>;
+}
```
tarcieri, line 92: These seem fairly similar to the [`EncodePrivateKey`](https://docs.rs/pkcs8/latest/pkcs8/trait.EncodePrivateKey.html) and [`EncodePublicKey`](https://docs.rs/pkcs8/latest/pkcs8/trait.EncodePublicKey.html) traits in the `pkcs8` crate.

Xynnn007: Yes. what I did is to simply wrap some functions in [EncodePrivateKey](https://docs.rs/pkcs8/latest/pkcs8/trait.EncodePrivateKey.html) and [EncodePublicKey](https://docs.rs/pkcs8/latest/pkcs8/trait.EncodePublicKey.html). 
The only different function here is `private_key_to_encrypted_pem()`. To be consistent with [sigstore's pem encrypt & decrypt](https://github.com/theupdateframework/go-tuf/blob/master/encrypted/encrypted.go) where `xsalsa20poly1305` was invloved, [`EncodePrivateKey`](https://docs.rs/pkcs8/latest/pkcs8/trait.EncodePrivateKey.html). The counterparts in `pkcs8` ,e.g. [to_pkcs8_encrypted_pem](https://docs.rs/pkcs8/latest/pkcs8/trait.EncodePrivateKey.html#method.to_pkcs8_encrypted_pem) doesn't support this scheme.

src/crypto/signing_key/ed25519.rs
Xynnn007: Yes, thanks! I've update related `from_pem` functions using `SecretDocument`

lukehinds:
```
---- cosign::verification_constraint::tests::cert_email_verifier_only_email stdout ----
thread 'cosign::verification_constraint::tests::cert_email_verifier_only_email' panicked at 'Cannot create certificate signature: CertificateValidityError("Not issued by a trusted root")', src/cosign/signature_layers.rs:558:18

---- cosign::verification_constraint::tests::cert_subject_url_verifier stdout ----
thread 'cosign::verification_constraint::tests::cert_subject_url_verifier' panicked at 'Cannot create certificate signature: CertificateValidityError("Not issued by a trusted root")', src/cosign/signature_layers.rs:558:18

---- cosign::verification_constraint::tests::pub_key_verifier stdout ----
thread 'cosign::verification_constraint::tests::pub_key_verifier' panicked at 'Cannot create certificate signature: CertificateValidityError("Not issued by a trusted root")', src/cosign/signature_layers.rs:558:18
```

@flavio is this to do with the TUF / tough root?

lukehinds:
I would really like to get this over the line, will try and help with the CI failures next week when I return from PTO.

Xynnn007:
> I would really like to get this over the line, will try and help with the CI failures next week when I return from PTO.

Thanks! And when `rsa` crate `v0.6.2` is published, I'll go on things on RSA module

tarcieri:
Note that the next release of `rsa` will probably be v0.7.0. We've merged some breaking changes.

src/crypto/signing_key/ed25519.rs
```diff
@@ -0,0 +1,358 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
cpanato, line 2: should be
```suggestion
// Copyright 2022 The Sigstore Authors.
```

src/crypto/signing_key/kdf.rs
```diff
@@ -0,0 +1,274 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
cpanato, line 2: should be
```suggestion
// Copyright 2022 The Sigstore Authors.
```

Cargo.toml
```diff
@@ -46,3 +58,4 @@ openssl = "0.10.38"
 rstest = "0.15.0"
 tempfile = "3.3.0"
 tracing-subscriber = { version = "0.3.9", features = ["env-filter"] }
+assert-json-diff = "2.0.2"
```
cpanato, line 61: nit: new line

src/crypto/signing_key/mod.rs
```diff
@@ -0,0 +1,244 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
cpanato, line 2: should be
```suggestion
// Copyright 2022 The Sigstore Authors.
```

src/crypto/signing_key/rsa.rs
```diff
@@ -0,0 +1,18 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
cpanato, line 2: should be
```suggestion
// Copyright 2022 The Sigstore Authors.
```

src/crypto/signing_key/ecdsa.rs
```diff
@@ -0,0 +1,432 @@
+//
+// Copyright 2021 The Sigstore Authors.
```
cpanato, line 2: should be
```suggestion
// Copyright 2022 The Sigstore Authors.
```

Xynnn007:
> @Xynnn007 can you please rebase your PR with the latest code from the main, that should fix the tests, thanks!

Sure, have done! And have fixed following your suggestions, thanks!

lukehinds:
Very close now @Xynnn007 , just some clippy stuff around Eq https://github.com/sigstore/sigstore-rs/runs/8018450036?check_suite_focus=true

cpanato:
PR here to fix the clippy lints issues https://github.com/sigstore/sigstore-rs/pull/98

lukehinds:
> PR here to fix the clippy lints issues #98

merged, do we need to rebase this PR or can we just re-run the clippy action?

cpanato:
@Xynnn007 please rebase when you have a chance :)

Xynnn007:
@cpanato Rebased, I was waiting for this :P

lukehinds (approved):

PR #99, opened 2022-08-27
feat: add example case and docs for key interface

Xynnn007:
#### Summary

As mentioned https://github.com/sigstore/sigstore-rs/issues/27, we need some examples for users to
use the key_interface module, s.t. `crypto/signing_key`. 

#### Release Note

* Added an example to use the `SigStoreSigner` in `crypto/signing_key`. Concretely, it Included a whole process of key pair generation, signing, exporting public/private key, importing public key and verifying the signature.
* Added a document for the example.

#### Documentation

None

### DIFF
--- a/examples/key_interface/README.md
+++ b/examples/key_interface/README.md
@@ -1,0 +1,27 @@
+# Example Key Interface
+
+This is a simple example program that shows how to perform signing key pair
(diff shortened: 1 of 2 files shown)


lukehinds (approved):
Thanks @Xynnn007

PR #111, opened 2022-08-31
Refactor examples to support subfolder execution

lukehinds:
To keep things more tidy, each example is moved under its core function
, for example `rekor` or `cosign`. A mapping is then made within
`Cargo.toml`.

This will make the folders more easily browsable for users and will allow
multiple file support. We can then use different file names beyond just
`main.rs`

```
[[example]]
name = "foo"
path = "examples/cosign/verify/foo.rs"

cargo run --example foo
```

```
[[example]]
name = "bar"
path = "examples/cosign/verify/bar.rs"

cargo run --example bar
```


lukehinds:
https://github.com/sigstore/sigstore-rs/pull/112 will address CI failure

cpanato (approved):

flavio (approved):
Thanks for the cleanup, LGTM

PR #112, opened 2022-08-31
Add RUSTSEC-2021-0139 to audit.toml

lukehinds:
Security CI runs are failing due to RUSTSEC-2021-0139

RUSTSEC-2021-0139 is flagged from ansi_term, which is a transient
dependency on tracing-subscriber. As tracing-subscriber is a
dev-dependency, I think we can safely add this to the ignore flag.

No vulnerabilities exist in ansi-term, the warning is there as the crate is
unmaintained at present.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -7,4 +7,7 @@ ignore = [
                        # to API breakages.
                        #
                        # This is a transitive depependency of tough
+  "RUSTSEC-2021-0139"  # ansi_term is no longer maintained, however this is a transient dependency of
+                       # the `tracing-subscriber` crate, which is a dev_dependency and so therefore
+                       # will not be included in a release build.
 ]


.cargo/audit.toml
```diff
@@ -7,4 +7,7 @@ ignore = [
                        # to API breakages.
                        #
                        # This is a transitive depependency of tough
+  "RUSTSEC-2021-0139"  # ansi_term is no longer maintained, however this is a transient dependency of
```
cpanato, line 10: this is something we can replace in the future? if yes can we create a tracking issue and mention here as well?

otherwise lgtm

cpanato (approved):

.cargo/audit.toml
lukehinds: It's not something we can replace as its a transient dependency through a different crate, but a crate that we use that has the transient is under development dependencies (so the code that is unmaintained won't is used in a release build). For that reason I don't think we need to be that concerned, but tracking won't hurt.

It was originally added back here: https://github.com/sigstore/sigstore-rs/commit/94a88595a83ba146df5a9f622c9911ebc1c58adc so @flavio might have a view of us being able to drop it entirely.

flavio:
This dev dependency is needed because we want our example programs to be able to print debug/warn/error messages to the standard output.

The `ansi-term` crate is being used by a lot of projects, hopefully this will lead to someone stepping up to be a maintainer or someone else just replacing it with another maintained crate.

In the meantime I think it's going to be simpler to get this PR merged.

PR #113, opened 2022-08-31
Update readme to include new features

lukehinds:
Has a soft dependency on https://github.com/sigstore/sigstore-rs/pull/111

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This is under high development, many features and checks are still missing.
 
 ## Features
 
-### Verification
+### CosignVerification
 
 The crate implements the following verification mechanisms:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,14 +23,43 @@ Signature annotations and certificate email can be provided at verification time
...
+
+### Key Interface
+
+The crate implements the following key interfaces:
+  
+    * Generate a key pair
+    * Sign data
+    * Verify signature
+    * Export key pair in PEM format
+    * Import key pair from PEM format
```
Xynnn007, line 38: Same, `Import (encrypted) private key in PEM format`.

README.md
```diff
@@ -23,14 +23,43 @@ Signature annotations and certificate email can be provided at verification time
...
+All of the rekor client APIs can be leveraged.
+
+### Key Interface
+
+The crate implements the following key interfaces:
+  
+    * Generate a key pair
+    * Sign data
+    * Verify signature
+    * Export key pair in PEM format
```
Xynnn007, line 37: I suppose here refers to `Export (encrypted) private key in PEM format`.

Xynnn007 (commented):
Also, do we need test for `Import/Export key in DER format`?

And I'll add these tests mentioned in this doc later.

README.md
lukehinds: Do we support DER format currently or are you recommending / asking if we should add it?

Does this look right:

    * Generate a key pair
    * Sign data
    * Verify signature
    * Export key pair in PEM / DER format
    * Import key pair from PEM / DER format

lukehinds: @Xynnn007 https://github.com/sigstore/sigstore-rs/pull/113/commits/e0f611bb56e9f60260ea2711d6600a372fb3f498

Xynnn007: We've already supported DER format now.

In fact we now exports either public key or private key, because the public key can be derived from the private key.
IMO
```
* Generate a key pair
* Sign data
* Verify signature
* Export public / (encrypted) private key in PEM / DER format
* Import public / (encrypted) private key in PEM / DER format
```

Besides, because now all the `KeyPair` trait objects are hided by an abstraction of `SigStoreSigner` for users to generate keypair due to the signing scheme easily, I think there should add more suitable interface for `SigStoreSigner` for the last `Import` test like `SigStoreSigner::from_private_der` and `SigStoreSigner::from_private_pem`. I wonder whether there is better design?

I'll take on this these days.

lkatalin (approved):

cpanato:
late lgtm :)

PR #115, opened 2022-08-31
feat: from and to interface for signing and verification keys

Xynnn007:
#### Summary

Improvements on https://github.com/sigstore/sigstore-rs/issues/27, for better interactive between `verification_key` module and `signing_key` module.

#### Release Note

- add `to_verification_key()` interface for `SigStoreSigner` and
`KeyPair` to derive `CosignVerificationKey`
- add `from_key_pair()` and `from_sigstore_signer()` interface for
`CosignVerificationKey`
- add unit tests for these new interfaces


Xynnn007:
Here the name is not `into_xxx` because the interface will keep the original object when generating a new object, rather than consume the origin object

src/crypto/signing_key/mod.rs
```diff
@@ -206,36 +231,36 @@ mod tests {
...
                 signing_scheme: SigningScheme::ED25519,
-                // This hash alg is not used, just as a placeholder.
-                verify_hash_alg: SignatureDigestAlgorithm::Sha256,
             },
         ];
 
         for item in test_items {
             let signer = SigStoreSigner::new(item.signing_scheme).unwrap();
-            let pubkey = signer.public_key_to_pem().unwrap();
+            let _pubkey = signer.public_key_to_pem().unwrap();
```
flavio, line 257: This is really a matter of personal taste, I prefer to use `expect` over these `.unwrap()` calls (like you do a couple of lines below) or, use something like:

```rust
assert(signer.public_key_to_pem().is_ok(), "cannot obtain public key of signer");
```

Feel free to leave the code as it is, this is really a matter of personal taste

flavio (approved):
LGTM, I left some minor comment that can be ignored

src/crypto/signing_key/mod.rs
lkatalin: I agree and think we will want to move away from `unwrap()`s sometime in the near future in favor of descriptive error messages. I think they are everywhere now so that can be its own Good First Issue.

lukehinds: +1. In fact CI should be failing instances of unwrap (will have to take a look).

src/crypto/signing_key/mod.rs
```diff
@@ -206,36 +231,36 @@ mod tests {
     /// * Randomly generate a key pair due to the given signing scheme.
     /// * Signing the MESSAGE and generate a signature using
     /// the private key.
+    /// * Derive the verification key using both `from_sigstore_signer`
+    /// and `to_verification_key`.
     /// * Verify the signature with the public key.
     #[test]
     fn sigstore_signing() {
```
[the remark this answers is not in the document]
Xynnn007: Thanks for introducing this crate. It is a pretty nice tool, much drier!

src/crypto/signing_key/mod.rs
Xynnn007: Yes, I agreed and replaced all the `unwrap()`s with `expect()`s showing reason and add extra notion for `assert`s. I think it is better when debugging and reading.

flavio (approved):
Thanks for the changes you have done, everything looks great to me

lukehinds:
Just a quick question, do we need a test to check verification using a cosign generated key, I know they use nacl secretbox there. @flavio you would have worked on this originally. I expect we are ok still.

Xynnn007:
> Just a quick question, do we need a test to check verification using a cosign generated key, I know they use nacl secretbox there. @flavio you would have worked on this originally. I expect we are ok still.

Hi, nice to see the question. In fact, [the key I used in the test](https://github.com/sigstore/sigstore-rs/blob/main/tests/data/keys/ecdsa_encrypted_private.key) is generated by cosign. And the relative logic for secretbox is in [this file](https://github.com/sigstore/sigstore-rs/blob/main/src/crypto/signing_key/kdf.rs). Now the compatibility with cosign-golang is pretty good.

lukehinds (approved):
nice work again @Xynnn007 , thanks for your contributions!

Issue #120, opened 2022-09-02
remove API for Rekor_Version endpoint


bobcallaway:
**Description**

https://github.com/sigstore/rekor/pull/1022 proposes that the `/api/v1/version` endpoint be removed from rekor, so opening this tracking issue to remove it from this client library.

PR #121, opened 2022-09-02
remove /api/v1/version from client

bobcallaway:
Fixes: #120 

Signed-off-by: Bob Callaway <bcallaway@google.com>


lkatalin (approved):

lukehinds:
Is the rekor version API deprecated? The code change looks good to me.

lukehinds (approved):
looks good to me. I think ultimately this code should be auto generated, but I know already some manual changes have been made. With that I think we should change as its then tracked.

PR #123, opened 2022-09-05
Refactoring and examples for key interface

Xynnn007:
#### Summary

Due to https://github.com/sigstore/sigstore-rs/issues/27, there is a need to refactor the key_interface to deal with the following current problems:
* When users use the `signing_key` module, they must import the crates into the `Cargo.toml`, like `p256`, `p384`, `signature`, `digest`etc. It is better for users not to need to care about the underlying Elliptic Curves. They can directly use `sigstore` crate without import any other crates to do all the key related operations.
* Now the module does not have a convenient call routine for a user to import a private key and convert it into a `SigStoreSigner`. Also `SigStoreSigner` is somehow overlapped functionally with `Signer` trait object. It is good to combine them, and give a more easy way for users to convert a PEM/DER encoded private key into an object who implements `sign`.

Thus this PR includes 
* Wrapped the generics trait object into `enum`s. 
* Combined `Signer` trait object and `SigStoreSigner` struct into an `enum` named `SigStoreSigner`. 
* Example and docs were attached.
#### Release Note

Added the new following APIs:
* `SigStoreKeyPair` `enum`
* `to_sigstore_signer()` for `SigStoreKeyPair`
* `to_sigstore_keypair()` for `SigStoreSigner`
* `ECDSAKeys` `enum`, a wrapper for `EcdsaKeys`

Changed the following `struct`s:
* Converted `SigStoreSigner` into an `enum`

Changed the following APIs:
* To create a new `SigStoreSigner`:`SigStoreSigner::new(SigningScheme::ECDSA_P256_SHA256_ASN1)` -> `SigningScheme::ECDSA_P256_SHA256_ASN1.create_signer()`

Added the following examples due to typical scenarios:
* Key Pair Generation, Signing and Verification
* Key Pair Generation and Exporting
* Key Pair Importing

Deleted the origin example:
* ecdsa_p256

To make it clear, an overview architecture of the implementation for key-interface was included.

#### Documentation

Please refer to the `key_interface` examples for the usages.


lukehinds:
Thanks @Xynnn007 , I do have this on my queue to review this week. Would be good to get @flavio to look over as well!

src/crypto/signing_key/ecdsa/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+
+use crate::errors::*;
+
+use self::ec::{EcdsaKeys, EcdsaSigner};
+
+use super::{KeyPair, SigStoreSigner};
+
+pub mod ec;
+
+pub enum ECDSAKeys {
```
flavio, line 29: nit, missing docs

src/crypto/signing_key/ed25519.rs
```diff
@@ -15,39 +15,21 @@
...
-//! let pubkey = ed25519_key_pair.public_key_to_pem().unwrap();
-//!
-//! // export the private key using sigstore encryption.
-//! let privkey = ed25519_key_pair.private_key_to_encrypted_pem(b"password").unwrap();
-//!
-//! // sign with the new key.
-//! let ed25519_signer = Ed25519Signer::from_ed25519_keys(&ed25519_key_pair).unwrap();
-//!
-//! let signature = ed25519_signer.sign(b"some message");
-//! ```
```
flavio, line 36: As before, it's a bit of a pity to drop examples from the documentation

src/crypto/signing_key/ecdsa/ec.rs
```diff
@@ -15,32 +15,6 @@
...
-//!
-//! // export the private key using sigstore encryption.
-//! let privkey = ec_key_pair.private_key_to_encrypted_pem(b"password").unwrap();
-//!
-//! // sign with the new key, using Sha256 as the digest scheme.
-//! // In fact, the signing scheme is ECDSA_P256_SHA256_ASN1 here.
-//! let ec_signer = EcdsaSigner::<_, sha2::Sha256>::from_ecdsa_keys(&ec_key_pair).unwrap();
-//!
-//! let signature = ec_signer.sign(b"some message");
-//! ```
```
flavio, line 43: It's a pity to remove examples from the inline docs, because this is what is visible on `docs.rs`. I know we have more examples under the `examples/` directory. I guess you did that to remove duplicated to code to maintain.

If that's the case, maybe we could experiment using [this feature](https://doc.rust-lang.org/rustdoc/scraped-examples.html?highlight=example#scraped-examples) of rustdocs to have one, or more, examples inlined into the user documentation :thinking:

flavio (approved):
LGTM, I left some minor comments

src/crypto/signing_key/ed25519.rs
Xynnn007: As mentioned, now added them. Thanks

src/crypto/signing_key/ecdsa/ec.rs
flavio: I see, that's a pity. We can definitely use the link to git :)

src/crypto/signing_key/ecdsa/mod.rs
flavio: Does it need to be public given the end users are not meant to use it?

Xynnn007: Now I just wrap `p256`, `p384` into the enum as they are usual ECs used for keypair and signing. However, there are also other curves, like `k256`, `p521`. If an "advanced user" want to use these curves, they can use the `ec` module for a low-level wrapper. How do you think about this?

flavio: That clarifies. thanks!

flavio (approved):

lukehinds (approved):

PR #127, opened 2022-09-13
Removed ring dependency

Xynnn007:
#### Summary

Due to https://github.com/sigstore/sigstore-rs/issues/27. This pr removed `ring` crate, and use RustCrypto crates instead.

#### Release Note

- replaced ring crate with RustCrypto
- refactored CosignVerificationKey implementation and interface
- fixed the code which references CosignVerificationKey
- added RSA keys and signing algorithms support (`RSA PSS` & `RSA PKCS#1 v1.5`)

##### API Changes

* Changed all the `from_XXX` functions for `CosignVerificationKey`'s `SignatureDigestAlgorithm` parameter into `SigningScheme`
* Deleted the enum `SignatureDigestAlgorithm`
* Replace struct `CosignVerificationKey` into enum.


src/crypto/verification_key.rs
```diff
@@ -13,94 +13,70 @@
...
-            };
-
-        let verification_algorithm = verification_algorithm?;
-        let data = public_key.subject_public_key.data.into_owned();
-
-        Ok(Self {
-            verification_algorithm,
-            data,
-        })
+        Self::from_der(key_pem.contents.as_slice(), signing_scheme)
```
flavio, line 79: I miss the usability of the old implementation. Now, as a user I would have to specify the signing scheme when importing the key. The signing scheme is made by encryption algorithm + digest algorithm. 

The old code leveraged a functionality of `ring` that provided back the type of encryption algorithm being used; see lines 69-95.

Thanks to that, the `verify` example was able to handle different kind of public keys without having the user to specify the encryption type.

Moreover, the current code lost the ability to handle RSA keys. I happen to have some container images that have been signed with a RSA key. This change breaks their verification.

In case you are interested:

```console
curl -O https://ftp.suse.com/pub/projects/security/keys/container–key.pem
cargo run --example verify -- -k ./container–key.pem registry.suse.com/suse/sle-micro/5.0/toolbox
```

This will fail with the following error:

```console
2022-09-13T12:21:07.095762Z  INFO sigstore::cosign::client_builder: Rekor public key not provided. Rekor integration disabled
2022-09-13T12:21:07.095804Z  INFO sigstore::cosign::client_builder: No Fulcio cert has been provided. Fulcio integration disabled
Image verification failed: Cannot create public key verifier: Pkcs8 spki error : Ecdsa-P256 from der bytes to public key failed: unknown/unsupported algorithm OID: 1.2.840.10045.2.1
```

flavio (commented):
Overall I like the change.

There are however two main things that have to be addressed:

* Be able to use RSA keys to verify signatures
* Given a PEM encoded key, have the code automatically figure out the encryption and signing algorithm. Or provide the user a helper function that does that

src/crypto/verification_key.rs
tarcieri: FWIW, the next release of the `rsa` crate will include some pretty major overhauls of signature support, including dedicated RSASSA-PSS and RSASSA-PKCS#1v1.5 signing/verifying keys which use the traits from the `signature` crate and hopefully support for things like parsing PSS params via PKCS#1/PKCS#8.

The current low-level signing APIs will also continue to work with some minor changes as well.

src/crypto/signing_key/ed25519.rs
```diff
@@ -399,16 +398,16 @@ mod tests {
         let pubkey = key
             .public_key_to_pem()
             .expect("export private key to PEM format failed.");
+        println!("{}", pubkey);
```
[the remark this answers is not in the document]
Xynnn007: Yea, sry for that. It was used for test.

src/crypto/verification_key.rs
Xynnn007: > I miss the usability of the old implementation. Now, as a user I would have to specify the signing scheme when importing the key. The signing scheme is made by encryption algorithm + digest algorithm.
> 
> The old code leveraged a functionality of `ring` that provided back the type of encryption algorithm being used; see lines 69-95.

In fact, the padding scheme used in the old implementation of RSA-related schemes only included `PSS`, so the API could only receive one parameter `DigestAlgorithm`. There are other padding schemes, like `OAEP`, `PKCS#1v1.5`. Thus EC keys may not share the same `from_pem/der` API with RSA keys.

Also, the public key file usually only contains information of the key itself. For example an EC public key only contains information of its type (ec public key, by OID) and curve (by OID), but not its signing usage.

So I think it is hard to do that, (TBH, I think the old implementation for the API is not reasonable)



> Moreover, the current code lost the ability to handle RSA keys. I happen to have some container images that have been signed with a RSA key. This change breaks their verification.

Maybe we can delay this pr. When the RSA key-interface is done, I'll rebase and go on doing this.

tarcieri: We're trying to wrap it up soon. This is probably the last major PR we'd like to get merged first prior to another release: https://github.com/RustCrypto/RSA/pull/183

Xynnn007: Ok, I got your idea about the developer convenience. I added two interfaces `try_from_der()` and `try_from_pem()`. The two interfaces receive only one parameter `data: &[u8]`, and will decide the verification algorithm (digest, padding ,...) with a default value. For example, by default `ECDSA_P256_SHA256_ASN1` for a `ec-p256 public key` and `RSA_PSS_2048_8192_SHA256` for a `RSA public key` (in future)

> * Go ahead with the `SignatureDigestAlgorithm` refactoring, they are pretty nice

What do you mean? Should `SignatureDigestAlgorithm` be kept or deleted? Sry I can not understand this ;)

flavio: > Ok, I got your idea about the developer convenience. I added two interfaces try_from_der() and try_from_pem(). The two interfaces receive only one parameter data: &[u8], and will decide the verification algorithm (digest, padding ,...) with a default value. For example, by default ECDSA_P256_SHA256_ASN1 for a ec-p256 public key and RSA_PSS_2048_8192_SHA256 for a RSA public key (in future)
Does this make sense?

Yes, it does make sense. I like the UX

> What do you mean? Should SignatureDigestAlgorithm be kept or deleted? Sry I can not understand this ;)

My fault, I like the cleaning/refactoring done by this PR. That's the way to go :+1:

flavio (commented):
I like the PR. If possible, I would like RSA support to be in before merging it. Otherwise, we can merge it in the main branch, but not tag a new release until this is added back.

src/crypto/verification_key.rs
Xynnn007: > I like the PR. If possible, I would like RSA support to be in before merging it. Otherwise, we can merge it in the main branch, but not tag a new release until this is added back.

Thanks. Let me finish RSA support in this PR then.

tarcieri: New `rsa` crate prerelease here:

https://docs.rs/rsa/0.7.0-rc.0/

It includes new PSS-specific `SigningKey`/`VerifyingKey` types:

https://docs.rs/rsa/0.7.0-rc.0/rsa/pss/struct.SigningKey.html

Xynnn007:
Please do not review the newly push-ed commit now, because it uses a modified version of `rsa` crate. I've made a PR for the change. After the PR is merged I'll change this commit.

lukehinds (approved):
great work thanks @Xynnn007

Xynnn007:
> great work thanks @Xynnn007

Hi @lukehinds  Thanks for your approval. 

However there is some significant things to clarify. This pr now involves **non-official** crypto crate version and **must not** be merged until the upstream of `rsa` merges the [pr](https://github.com/RustCrypto/RSA/pull/191).

I pushed the [second commit](https://github.com/sigstore/sigstore-rs/pull/127/commits/460de25c8e39e7d7e7ede5b6ff39dde41366d755) to collect suggestions about:
* Whether it is rational to add a member `usize` for RSA-related signing schemes enums to specify the key length. Or we set a default value of the key size sacrifice the flexibility for avoidance for users to set it manually (like 2048 or 4096 bits)

lukehinds:
Got you @Xynnn007 you have the dependency pinned to your fork until upstream changes merge. Thanks for pointing that out, I will change this to a hold.

Regarding the two options. I am more inclined to allow users to pass a `usize` in. How about we allow users to set key length, yet if the key length is None, we default to 4096?

tarcieri:
Note: the RSA patch is merged

Xynnn007:
> Note: the RSA patch is merged

Great, thanks! When the new version of `rsa` is published I will import it, instead of `{git = "xxx", rev = "xxx"}`

lukehinds:
@tarcieri do you have any insight into when v0.7.0 moves from rc to a release?

lukehinds:
> We can already consume the rc from `crates.io`: https://crates.io/crates/rsa/0.7.0-rc.0

Is this sufficient @Xynnn007 ?

flavio (dismissed):

flavio:
I tried building using the released crate and the code doesn't compile.

Aside from that, I tried the current code (the one that builds)::

```console
curl -O https://ftp.suse.com/pub/projects/security/keys/container–key.pem
cargo run --example verify -- -k ./container–key.pem registry.suse.com/suse/sle-micro/5.0/toolbox
```

and the verification fails. I didn't look too much into that, maybe @Xynnn007 has a better idea about why this RSA key is not working anymore :thinking:

Xynnn007:
Yea, I pushed a pr on `0.7.0-rc.0` to help, so we should wait for the newest tag after `0.7.0-rc.0`
@lukehinds @flavio

tarcieri:
@lukehinds we'd like to get a final release out soon, although it'd be good to know if there are any remaining issues.

lukehinds:
I think we should merge this, and hold on the release for 0.7.0 , at which point we can just update cargo.toml, that way it will save folks managing rebases when this later lands

How does that sound @Xynnn007 ?

Xynnn007:
> I think we should merge this, and hold on the release for 0.7.0 , at which point we can just update cargo.toml, that way it will save folks managing rebases when this later lands. We can cut the release then.
> 
> How does that sound @Xynnn007 ?

Ok, I've just change the crate ref. Also, current implementation may not be perfect as I see, and we can make it better over time.

lukehinds (dismissed):

flavio:
Please, before merging it, can you figure out what is wrong with the RSA signature I've pointed out [here](https://github.com/sigstore/sigstore-rs/pull/127#issuecomment-1262205285)?

This is a regression

Xynnn007:
> @Xynnn007 Please, before merging it, can you figure out what is wrong with the RSA signature I've pointed out [here](https://github.com/sigstore/sigstore-rs/pull/127#issuecomment-1262205285)?
> 
> This is a regression

Ah, please wait for some time. Let me do this

flavio:
> Ah, please wait for some time. Let me do this

Sure, no hurries! :smile: 

I just wanted to ensure this PR wouldn't get accidentally merged with this open issue.

Xynnn007:
@flavio Hey, there is a trick. The image you've given is sign by `RSA_PKCS1_SHA256`, and I set the default signing scheme of RSA is `RSA_PSS_SHA256`, because the padding scheme `PSS` is safer than `PKCS1`

You can use an extra parameter to tell the signing scheme
```bash
cargo run --example verify -- -k ./container–key.pem --signing-scheme RSA_PKCS1_SHA256 registry.suse.com/suse/sle-micro/5.0/toolbox
```

And will result
```
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/examples/verify -k './container–key.pem' --signing-scheme RSA_PKCS1_SHA256 registry.suse.com/suse/sle-micro/5.0/toolbox`
2022-09-30T12:08:50.238132Z  INFO sigstore::cosign::client_builder: Rekor public key not provided. Rekor integration disabled
2022-09-30T12:08:50.238193Z  INFO sigstore::cosign::client_builder: No Fulcio cert has been provided. Fulcio integration disabled
Image successfully verified
```

Still hold on, please. I found some bugs for notes, let me fix it in a few minutes.

Xynnn007:
Done, please take a look whether there is problems left
@flavio @lukehinds

lukehinds:
could you rebase with https://github.com/sigstore/sigstore-rs/pull/136 pls

Xynnn007:
@lukehinds hey done

lukehinds:
thanks @Xynnn007 

@flavio is your comment still applicable?

lukehinds:
I should have held off 0.5.0 for this one, my bad. @flavio as above, are you OK with this one to merge?

flavio:
> @flavio Hey, there is a trick. The image you've given is sign by `RSA_PKCS1_SHA256`, and I set the default signing scheme of RSA is `RSA_PSS_SHA256`, because the padding scheme `PSS` is safer than `PKCS1`
> 
> You can use an extra parameter to tell the signing scheme
> 
> ```shell
> cargo run --example verify -- -k ./container–key.pem --signing-scheme RSA_PKCS1_SHA256 registry.suse.com/suse/sle-micro/5.0/toolbox
> ```
> 
> And will result
> 
> ```
>     Finished dev [unoptimized + debuginfo] target(s) in 0.14s
>      Running `target/debug/examples/verify -k './container–key.pem' --signing-scheme RSA_PKCS1_SHA256 registry.suse.com/suse/sle-micro/5.0/toolbox`
> 2022-09-30T12:08:50.238132Z  INFO sigstore::cosign::client_builder: Rekor public key not provided. Rekor integration disabled
> 2022-09-30T12:08:50.238193Z  INFO sigstore::cosign::client_builder: No Fulcio cert has been provided. Fulcio integration disabled
> Image successfully verified
> ```

Please, forgive me if I bring this topic up again. Is there a way to understand if a RSA key is using the pkcs v1 1.5 or the pss signature algorithm? I tried searching for it, but I couldn't find anything about that. Maybe @Xynnn007 or @tarcieri know about that...

I'm so obsessed by this topic because I would like the consumers of the `sigstore` crate to be able to build tools that behave like cosign. With cosign, I don't have to specify what kind of algorithms are used by an input key, it will try to guess it using a list of supported algorithms.

Now, this PR is doing something in this direction too with the `PublicKeyVerifier::try_from` method. It would be great to extend this method to cover also RSA_PKCS1 verification keys.

tarcieri:
> Is there a way to understand if a RSA key is using the pkcs v1 1.5 or the pss signature algorithm? 

I'm not 100% clear on this, but I believe PKCS#1 v2.2 as defined in RFC8017 makes it possible to mark a key as intended for use with PSS, and also provide the additional PSS parameters (e.g. hash function, MGF, and salt length) needed to instantiate the algorithm. See:

https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.2.3

tarcieri:
I released `rsa` v0.7.0-rc.1 which may fix the compile errors.

We also have a tracking issue for a final v0.7.0 release here: https://github.com/RustCrypto/RSA/issues/205

flavio:
> I released `rsa` v0.7.0-rc.1 which may fix the compile errors.

Yes, that fixes the compilation errors. Thanks!

I've updated this PR to make use of this pre-release from crates.io instead of obtaining the code from the git repository

src/crypto/verification_key.rs
```diff
@@ -13,94 +13,179 @@
...
+    /// set the verification algorithm due to the public key type, s.t.
+    /// * `RSA public key`: `RSA_PSS_SHA256`
+    /// * `EC public key with P-256 curve`: `ECDSA_P256_SHA256_ASN1`
+    /// * `EC public key with P-384 curve`: `ECDSA_P384_SHA384_ASN1`
+    /// * `Ed25519 public key`: `Ed25519`
+    pub fn try_from_der(der_data: &[u8]) -> Result<Self> {
+        if let Ok(p256vk) = ecdsa::VerifyingKey::from_public_key_der(der_data) {
+            Ok(Self::ECDSA_P256_SHA256_ASN1(p256vk))
+        } else if let Ok(p384vk) = ecdsa::VerifyingKey::from_public_key_der(der_data) {
+            Ok(Self::ECDSA_P384_SHA384_ASN1(p384vk))
```
flavio, line 156: I just noticed these lines of code. I don't understand how the same method `ecdsa::VerifyingKey::from_public_key_der` could lead to identify a `ECDSA_P256_SHA256_ASN1` and a `ECDSA_P384_SHA384_ASN1`.

To put it in another way... I don't understand how the code can identify these two types of keys.

Maybe I'm missing something obvious :sweat_smile:

Xynnn007: Here the der-bytes of public key will follow `SPKI(SubjectPublicKeyIdentifier)` format, which will contain `P384` curve and `P256` curve information. The two functions are generic functions and concrete type will be inferred by compiler, s.t.
```rust
if let Ok(p256vk) = ecdsa::VerifyingKey::<p256::NistP256>::from_public_key_der(der_data) {
            Ok(Self::ECDSA_P256_SHA256_ASN1(p256vk))
        } else if let Ok(p384vk) = ecdsa::VerifyingKey::<p384::NistP384>::from_public_key_der(der_data) {
            Ok(Self::ECDSA_P384_SHA384_ASN1(p384vk))
```

flavio: :facepalm: sorry... I had a brain fart and I missed that

flavio (approved):
Thanks a lot @Xynnn007 for the awesome work and for your patience during the review process! :clap: :heart:

tarcieri:
FYI `rsa` v0.7.0 has been released: https://crates.io/crates/rsa/0.7.0

src/crypto/verification_key.rs
tarcieri: FYI, for support for PKCS#1/PKCS#8 PSS keys, this is a tracking issue specifically for supporting them:

https://github.com/RustCrypto/RSA/issues/177

Xynnn007:
> FYI `rsa` v0.7.0 has been released: https://crates.io/crates/rsa/0.7.0

Yea, thanks!

Issue #129, opened 2022-09-20
Provide pure async variants for OIDC flows.


bobmcwhirter:
Current implementation of the OIDC flows requires blocking tasks.

This is limiting when applications are otherwise using `awc` which leads to using `actix_rt` for async executors.

Providing `async_http_client` variants for `auth_url()` and `redirect_listener()` would remove this limitation.

**note** this issue provided for tracking on a PR I'm opening.

jvanz:
Considering that the PR of this issue is closed. I think this issue can be closed as well.

flavio:
Closing, we now have async implementations of both `auth_url` and `redirect_listener`

PR #130, opened 2022-09-20
Add _async variants of methods using async http client for OIDC flow.

bobmcwhirter:
This supports usage without spawning blocking tasks, at least with actix_rt.

Fix for https://github.com/sigstore/sigstore-rs/issues/129

#### Summary

When using actix_rt, the blocking variants for OIDC flows cause drop panics.  Providing async variants allows straightforward usage within an actix_rt-centric application.

#### Release Note

* Async variations of `auth_url()` and `redirect_listener()` with `_async` suffixes.

### DIFF
--- a/src/oauth/openidflow.rs
+++ b/src/oauth/openidflow.rs
@@ -87,7 +87,7 @@ use openidconnect::core::{
     CoreClient, CoreIdToken, CoreIdTokenClaims, CoreIdTokenVerifier, CoreProviderMetadata,
     CoreResponseType,
 };
-use openidconnect::reqwest::http_client;
+use openidconnect::reqwest::{async_http_client, http_client};
 use openidconnect::{
     AuthenticationFlow, AuthorizationCode, ClientId, ClientSecret, CsrfToken, IssuerUrl, Nonce,
     PkceCodeChallenge, PkceCodeVerifier, RedirectUrl, Scope,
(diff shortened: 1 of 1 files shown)


src/oauth/openidflow.rs
```diff
@@ -158,6 +158,34 @@ impl OpenIDAuthorize {
...
         Ok((authorize_url, client, nonce, pkce_verifier))
     }
+
+    pub async fn auth_url_async(&self) -> Result<(Url, CoreClient, Nonce, PkceCodeVerifier)> {
+        let client_id = ClientId::new(self.oidc_cliend_id.to_owned());
+        let client_secret = ClientSecret::new(self.oidc_client_secret.to_owned());
+        let issuer = IssuerUrl::new(self.oidc_issuer.to_owned()).expect("Missing the OIDC_ISSUER.");
+
+        let (pkce_challenge, pkce_verifier) = PkceCodeChallenge::new_random_sha256();
+        let provider_metadata = CoreProviderMetadata::discover_async(issuer, async_http_client)
```
lukehinds, line 168: I wonder if we could use a match option here, to save duplicating the whole function. Just trying to think of ways to prevent DRY here.

lukehinds (commented):
Thanks, agree it would be useful to get async in.

src/oauth/openidflow.rs
bobmcwhirter: I think function coloring between async/non-async will jam us up, but I'm sure there's some way to reduce copy pasta.

Issue #131, opened 2022-09-20
Fulcio client support


bobmcwhirter:
Provide a Fulcio client capable of creating and exchanging keys for certificates.

Inbound PR shortly.

PR #132, opened 2022-09-20
Fulcio client

bobmcwhirter:
Fixes https://github.com/sigstore/sigstore-rs/issues/131

#### Summary

Creates an async Fulcio client to make working with Fulcio more straightfoward.

#### Release Note
* Fulcio client for exchanging public key for signed certificate added.
 
#### Documentation

Rust docs provided.

### DIFF
--- a/src/fulcio/mod.rs
+++ b/src/fulcio/mod.rs
@@ -1,0 +1,154 @@
+use crate::crypto::signing_key::{SigStoreSigner, SigningScheme};
+use crate::errors::{Result, SigstoreError};
+use openidconnect::core::CoreIdToken;
+use openidconnect::reqwest::async_http_client;
+use reqwest::header::HeaderName;
+use reqwest::Body;
+use serde::ser::SerializeStruct;
+use serde::{Deserialize, Serialize, Serializer};
+use std::convert::{TryFrom, TryInto};
+use std::fmt::{Debug, Display, Formatter, Write};
+use std::io::Read;
+use std::str::from_utf8;
+use url::Url;
+use x509_parser::nom::AsBytes;
+
+/// Default public Fulcio server root.
+pub const FULCIO_ROOT: &'static str = "https://fulcio.sigstore.dev/";
+
+/// Path within Fulcio to obtain a signing certificate.
+pub const SIGNING_CERT_PATH: &str = "api/v1/signingCert";
+
+const CONTENT_TYPE: HeaderName = HeaderName::from_static("content-type");
+
+/// Fulcio certificate signing request
+///
+/// Used to present a public key and signed challenge/proof-of-key in exchange
+/// for a signed X509 certificate in return.
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+struct Csr {
+    public_key: Option<PublicKey>,
+    signed_email_address: Option<String>,
+}
+
+impl TryFrom<Csr> for Body {
+    type Error = serde_json::Error;
+
+    fn try_from(csr: Csr) -> std::result::Result<Self, Self::Error> {
+        Ok(Body::from(serde_json::to_string(&csr)?))
+    }
+}
+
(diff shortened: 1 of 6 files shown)


bobmcwhirter:
I want to rework this a bit I think.

lukehinds:
This might be better suited for `examples/` , it would be very valuable as a guide. In examples you can use unwrap as much as you like as well. Happy to see what other reviewers think though.

bobmcwhirter:
I do still think expanding it and including it in the mainline would be useful,  particularly in light of the protobuf-vs-json-vs-whatever discussions. By keeping it in the main tree, downstream users won't have to individually adapt to protocol changes.

I do want to make it all Much Better. Ignore this for now (I'll mark it as a draft) and continue to iterate.

bobmcwhirter:
btw, happy for a review now.

src/oauth/openidflow.rs
```diff
@@ -241,33 +262,66 @@ impl RedirectListener {
...
+            .expect("Server did not return an ID token")
+            .claims(&id_token_verifier, &self.nonce)
+            .map_err(|err| {
+                error!("Error is: {:?}", err);
+                SigstoreError::ClaimsVerificationError
+            })?;
+        Ok((id_token_claims.clone(), id_token.clone()))
+    }
+
+    pub async fn redirect_listener_async(self) -> Result<(CoreIdTokenClaims, CoreIdToken)> {
```
lukehinds, line 300: Could we could combine `redirect_listener_async` and `redirect_listener`, as they are largely the same apart from a single line (calling either `http_client` or `async_http_client`

src/oauth/openidflow.rs
```diff
@@ -271,55 +271,57 @@ impl RedirectListener {
...
+            pkce_verifier,
+            nonce,
+            client,
+            ..
+        } = self;
+
+        let token_response = client
             .exchange_code(code)
-            .set_pkce_verifier(self.pkce_verifier)
+            .set_pkce_verifier(pkce_verifier)
```
flavio, line 283: nit, this is a matter of personal taste. I prefer to use `self.<attribute>` instead of "unpacking" the contents of `self`. That makes easier to understand where a certain value is coming from (like `pkce_verifier`).

bobmcwhirter:
I agree but this due to borrow issues.

src/oauth/openidflow.rs
bobmcwhirter: Apparently I was just having a brain fart.

I've removed the destructuring of Self now.

I've also rebased against `main#HEAD`

lukehinds (commented):
Just one more spotted, sorry about the slow reviews from me @bobmcwhirter

src/fulcio/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+use reqwest::header::HeaderName;
+use reqwest::Body;
+use serde::ser::SerializeStruct;
+use serde::{Serialize, Serializer};
+use std::convert::{TryFrom, TryInto};
+use std::fmt::{Debug, Display, Formatter};
+use url::Url;
+
+/// Default public Fulcio server root.
+pub const FULCIO_ROOT: &str = "https://fulcio.sigstore.dev/";
```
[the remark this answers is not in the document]
bobmcwhirter: The client takes a URL in the ctor. Const just handy for folks who want to use public fulcio.

lukehinds (approved):
Just a couple of fmt changes to make, but lgtm!

src/oauth/openidflow.rs
flavio: thanks!

flavio (commented):
Please fix the rustfmt errors, then we can go ahead and merge it

bobmcwhirter:
> @bobmcwhirter Please fix the rustfmt errors, then we can go ahead and merge it. I'm good with it too

Done!

flavio (approved):
thanks for the last round of fixes

PR #133, opened 2022-09-21
Implement cargo vet for improved package security

lukehinds:
cargo vet from mozilla allows us to benefit from crate audits performed by other communities such as mozilla, bytecodealliance etc.

This PR sets a baseline which will pass. If we wish to, we can audit back at any time. The changes in this PR will (once merged) fail a PR if a new package is introduced with no audit history from our peer network set within `supply-chain/audits.toml`.

All we then need to do, is perform a review of the crate and decide if we feel it is safe to include into sigstore-rs

Anyone can pre-audit with for example a git hook, by installing cargo vet `cargo install cargo-vet` and then running an audit: `cargo vet`.

This is useful for obvious reasons, and could capture us mistakenly using a typo based attack or introducing an unaudited crypto lib.

Last of all, we bring Cargo.lock back in, which should only really be gitignore'd for a binary.


.github/workflows/cargo-vet.yml
```diff
@@ -0,0 +1,22 @@
+name: cargo-vet
+on: [push, pull_request]
+jobs:
+  cargo-vet:
+    name: Vet Dependencies
+    runs-on: ubuntu-latest
+    env:
+      CARGO_VET_VERSION: 0.3.0
+    steps:
+    - uses: actions/checkout@master
```
cpanato, line 10: we avoid to use the main branch and also tags, can you please update this to 
```
actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
```

.github/workflows/cargo-vet.yml
```diff
@@ -0,0 +1,22 @@
...
+    - name: Install Rust
+      run: rustup update stable && rustup default stable
+    - uses: actions/cache@v2
```
cpanato, line 13: same here 
```suggestion
    - uses: actions/cache@fd5de65bc895cf536527842281bea11763fefd77 # v3.0.8
```

lukehinds:
hold on merging this, just found https://github.com/crev-dev/cargo-crev and figuring out which is better suited

cpanato (changes requested):
small change request

otherwise lgtm :)

.github/workflows/cargo-vet.yml
```diff
@@ -0,0 +1,22 @@
+name: cargo-vet
+on: [push, pull_request]
```
cpanato, line 2: here will be good to lock in the main branch

```
on:
  push:
    branches:
      - main
  pull_request:
  workflow_dispatch:
```


and maybe add the workflow_dispatch in case we want to run manually

flavio:
> Last of all, we bring Cargo.lock back in, which should only really be gitignore'd for a binary.

Committing the `Cargo.lock` is not recommended for libraries. I guess we have to do that to ensure `cargo vet` information are accurate, I just wonder if that could have side effects when consuming the library :thinking:

lukehinds:
> > Last of all, we bring Cargo.lock back in, which should only really be gitignore'd for a binary.
> 
> Committing the `Cargo.lock` is not recommended for libraries. I guess we have to do that to ensure `cargo vet` information are accurate, I just wonder if that could have side effects when consuming the library 🤔

You're right, i got it backwards.

Issue #135, opened 2022-09-29
Verify not working if the certificate annotation is present and there is no bundle annotation


raulcabello:
**Description**

Both certificate and bundle are [optional parameters](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md#properties) when signing with cosign. If the certificate annotation is present and there is no bundle annotation, verification fails with the following error:

```
2022-09-29T14:56:46.298309Z  INFO sigstore::cosign::signature_layers: Skipping OCI layer because of error error=SigstoreRekorBundleNotFoundError
Image verification failed: No Signature Layer passed verification
```

This happens when providing a certificate when signing with `cosign` using a keypair and without enabling `COSIGN_EXPERIMENTAL`

**How to reproduce it?**

`raulcabm/sigstore-test:1` was signed with cosign and [pkcs11](https://github.com/sigstore/cosign/blob/main/PKCS11.md). `dev.cosignproject.cosign/signature` and `dev.sigstore.cosign/certificate` annotations are present, which is a valid scenario. 

`crane manifest $(cosign triangulate  raulcabm/sigstore-test:1) | jq`

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "size": 248,
    "digest": "sha256:2d4f0c0130a23ca7ab0aa545aadb91674ba7eef0bb38ebada113b047d6508bad"
  },
  "layers": [
    {
      "mediaType": "application/vnd.dev.cosign.simplesigning.v1+json",
      "size": 254,
      "digest": "sha256:25fa425dace3302722778703a6b4387251af7c3a8513827f05010c0d0785d9f4",
      "annotations": {
[...]

PR #142, opened 2022-10-12
fix verification of signatures produced with pki11

flavio:
The `cosign` tool can produce signatures using a PKCS11 token. These signatures feature a `certificate` annotation inside of their OCI layer.

However, when `COSIGN_EXPERIMENTAL` is not enabled, the layer will not feature a Rekor bundle.

Prior to this commit, the code assumed signature layers could have a `certificate` annotation only when using the Fulcio integration. Because of that, layers with a `certificate` but without a Rekor bundle were discarded. That was done to ensure the robustness of keyless verification.

This commit changes the code that creates `SignatureLayer` objects to not raise errors when an embedded certificate cannot be verified. Be it because it has been forged/invalid/etc or because the Rekor bundle is not found inside of the layer.

The resulting `SignatureLayer` will not be discarded, but it will have its `certificate_signature` attribute set to `None`.

> **Note:** `SignatureLayer::certificate_signature` was already a `Option` before of this commit.

The verification constraints implementing keyless verification will discard these kind of layers because they do not have a `certificate_signature`.

However, the public key based verifier will be able to verify the signature stored inside of the layer.

This solves the following scenario:

> Given Alice signed a container image using a PKCS11 token but without having cosign's Rekor integration enabled
> When verifying the container image signature using the public key associated with the certificate stored on her PKCS11 token
> Then the container image will be reported as successfully verified

Fixes https://github.com/sigstore/sigstore-rs/issues/135

### DIFF
--- a/src/cosign/signature_layers.rs
+++ b/src/cosign/signature_layers.rs
@@ -116,7 +116,24 @@ pub struct SignatureLayer {
     /// The digest of the layer
     pub oci_digest: String,
     /// The certificate holding the identity of the signer, plus his
-    /// verification key. This exists for signature done with keyless mode.
+    /// verification key. This exists for signature done with keyless mode or
+    /// when a PKCS11 token was used.
+    ///
+    /// The value of `CertificateSignature` is `None`
+    /// when no certificate was embedded into the
+    /// layer, or when the embedded certificate could not be verified.
+    ///
+    /// Having a `None` value will rightfully cause the
+    /// keyless verifiers like
+    /// [`CertSubjectEmailVerifier`](crate::cosign::verification_constraint::CertSubjectEmailVerifier)
+    /// or
+    /// [`CertSubjectUrlVerifier`](crate::cosign::verification_constraint::CertSubjectUrlVerifier)
+    /// to fail verification.
+    /// However, it will still be possible to use the
+    /// [`PublicKeyVerifier`](crate::cosign::verification_constraint::PublicKeyVerifier)
+    /// to verify the layer. This can be useful to verify signatures produced
+    /// with a PKCS11 token, but with Rekor's integration disabled at
+    /// signature time.
     pub certificate_signature: Option<CertificateSignature>,
     /// The bundle produced by Rekor.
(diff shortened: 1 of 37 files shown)


src/cosign/signature_layers.rs
```diff
@@ -751,17 +776,12 @@ JsB89BPhZYch0U0hKANx5TY+ncrm0s8bfJxxHoenAEFhwhuXeb4PqIrtoQ==
...
             Some(&bundle),
-        )
-        .expect_err("Didn't get an error");
-
-        assert!(matches!(
-            error,
-            SigstoreError::SigstoreFulcioCertificatesNotProvidedError
-        ));
+        );
+        assert!(cert.is_none());
```
viccuad, line 784: I think the test functions should be renamed now, e.g: 
```
get_certificate_signature_from_annotations_fails_when_no_bundle_is_given()
get_certificate_signature_from_annotations_fails_when_no_fulcio_pub_key_is_given()
```
should be renamed to `[...]_returns_none()`

viccuad (approved):
LGTM! just a minor nit.

raulcabello:
I like this approach. This PR fix https://github.com/sigstore/sigstore-rs/issues/135 by not failing if the certificate in the `dev.sigstore.cosign/certificate` annotation is present when there is no rekor bundle, which is a valid scenario as explained [above](https://github.com/sigstore/sigstore-rs/pull/142#issue-1406425664). 
Public key will still be verified in the [PublicKeyVerifier](https://github.com/sigstore/sigstore-rs/blob/main/src/cosign/verification_constraint.rs#L103). 
Keyless verification will keep working as expected as rekor bundle should be present. If it is not present verification will fail because the `certificate_signature` will be `None` in the [CertSubjectEmailVerifier](https://github.com/sigstore/sigstore-rs/blob/main/src/cosign/verification_constraint.rs#L218) or [CertSubjectUrlVerifier](https://github.com/sigstore/sigstore-rs/blob/main/src/cosign/verification_constraint.rs#L283)

PR #146, opened 2022-10-20
fix: rustls should not require openssl

flavio:
Ensure the openssl dependency is not introduced when the sigstore-rs library is built using rustls-tls.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "sigstore"
 description = "An experimental crate to interact with sigstore"
-version = "0.5.2"
+version = "0.5.3"
 edition = "2018"
 authors = [
   "sigstore-rs developers",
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -11,15 +11,15 @@ readme = "README.md"
 
 [features]
 default = ["native-tls"]
-native-tls = ["oci-distribution/native-tls", "openidconnect/native-tls"]
-rustls-tls = ["oci-distribution/rustls-tls", "openidconnect/rustls-tls"]
+native-tls = ["oci-distribution/native-tls", "openidconnect/native-tls", "openidconnect/native-tls", "reqwest/native-tls"]
+rustls-tls = ["oci-distribution/rustls-tls", "openidconnect/rustls-tls", "openidconnect/rustls-tls", "reqwest/rustls-tls"]
```
viccuad, line 15: we are repeating `"openidconnect/foo"` in both features without gain.

flavio: you're right!

viccuad (approved):
LGTM, thanks!

PR #147, opened 2022-10-20
Add getter functions for LogEntry fields

lkatalin:
Signed-off-by: Lily Sturmann <lsturman@redhat.com>

#### Summary
I need to access some of these fields, and I think being able to do so would make sense.


lkatalin:
Related to #150

flavio (approved):
Maybe we could simply make these fields public, to avoid having to write all these getters. This looks a bit Java style, but I have nothing against that

lkatalin:
> Maybe we could simply make these fields public, to avoid having to write all these getters. This looks a bit Java style, but I have nothing against that

Thanks for the review! Either works for me. Sometimes I think the getters are more flexible, but in this case maybe you are right since the retrieval is so simple.

lkatalin:
@flavio I rewrote it.

lukehinds (approved):
lgtm!

PR #152, opened 2022-10-24
Updates for parsing hashedrekord LogEntry

lkatalin:
#### Summary
Updates to remove nonexistent fields and add functionality to parse hashedrekord LogEntry struct.


src/rekor/models/hashedrekord.rs
```diff
@@ -79,18 +79,23 @@ impl PublicKey {
     pub fn new(content: String) -> PublicKey {
         PublicKey { content }
     }
+
+    pub fn decode(&self) -> Result<String, SigstoreError> {
+        let decoded = decode(&self.content)?;
+        let decoded_str = std::str::from_utf8(&decoded)?;
+        Ok(String::from(decoded_str))
```
flavio, line 86: nitpicking, this can replaced with something like `String::from_utf8`

src/rekor/models/log_entry.rs
```diff
@@ -14,6 +18,25 @@ pub struct LogEntry {
     pub verification: Verification,
 }
 
+impl LogEntry {
+    pub fn decode_body(&self) -> Result<Body, SigstoreError> {
+        let decoded = decode(&self.body)?;
+        let decoded_str = std::str::from_utf8(&decoded)?;
+        let body = serde_json::from_str::<Body>(decoded_str)?;
```
flavio, line 25: I think this could be replaced with:

```rust
let body = serde_json::from_slice(decoded)?;
```

lkatalin:
You are right, thank you @flavio!

src/rekor/models/hashedrekord.rs
```diff
@@ -79,18 +79,22 @@ impl PublicKey {
     pub fn new(content: String) -> PublicKey {
         PublicKey { content }
     }
+
+    pub fn decode(&self) -> Result<String, SigstoreError> {
+        let decoded = decode(&self.content)?;
+        String::from_utf8(decoded).map_err(|e| SigstoreError::from(e.utf8_error()))
```
flavio, line 85: This is totally fine, but if you want to save some keystrokes (and avoid to go crazy figuring out how to convert the error), I find totally acceptable to do something like that:

```rust
let response = String::from_utf8(decoded)?;
Ok(response)
```

flavio (approved):
LGTM

src/rekor/models/hashedrekord.rs
flavio: let me know if you want me to merge this as it is, or if you want to use the code from above. I'm totally fine with the current PR

lkatalin: Thanks, I was going a little bit crazy. :upside_down_face: However, I just tried this and it gives me an error
```
error[E0277]: `?` couldn't convert the error to `errors::SigstoreError`
  --> src/rekor/models/hashedrekord.rs:86:50
   |
86 |         let response = String::from_utf8(decoded)?;
   |                                                  ^ the trait `From<FromUtf8Error>` is not implemented for `errors::SigstoreError`

```

I can always implement `FromUtf8Error` as a `SigstoreError` in `errors.rs`, but it is so similar to `Utf8Error` that it seemed redundant. Is there still a way to use your trick in this case?

flavio: ouch, yes... that would be redundant.

I think you could map this error into a more generic Sigstore one, like the `UnexpectedError`, you could build a message string with `format!` and provide some more context about the failure

lkatalin: Hmm, yes, that is an option. I think mapping it to `Unexpected` with a string format would be more verbose (and in a sense less precise). Would you want to keep it as-is since we can't automatically convert to `SigstoreError` cleanly? Unless you think we do need more context, then I can add some.

flavio: let's leave as it currently is, it's much cleaner. If we need to map this type of error again, we can then implement a `StringFromUf8Error` or something like that inside of `SigstoreError`

PR #153, opened 2022-10-24
Add functionality to parse pubkey from Fulcio cert

lkatalin:
Signed-off-by: Lily Sturmann <lsturman@redhat.com>


flavio (changes requested):
I'm against that, because this would cause openssl to become a runtime dependency, making impossible for consumers of this library to build their libraries/binaries as pure Rust.

If you want I can help with the rewrite of this code using the crates that are already being consumed.

BTW: currently `openssl` is a development dependency because it's used to create x509 certificates inside of the unit tests. This code could be rewritten using `picky-rs`, but I didn't consider that worth the effort.

lukehinds:
I had not noticed that, incorrectly read it as a version bump. Agree with @flavio here, a lot of minimal images ship without openssl.

lkatalin:
Sure, I can rewrite it with one of the existing crates probably after this week... stay tuned!

flavio:
Thanks a lot @lkatalin for understanding the issue

Issue #157, opened 2022-11-10
Add ability to verify signatures with a user provided certificate


flavio:
**Description**

cosign allows users to sign objects using a certificate that has been issued by a user-managed root CA.

Consider the following scenario:

> Company Acme Inc. runs its own PKI. An intermediate CA is used to issue certificates to its employees.
> Employees rely on these certificates to sign their container images using cosign.

In this scenario, as a sigstore-rs consumer, I would like to be able to verify a container image by using the following input data:

* user certificate
* certificate chain
* name of the object to be verified

flavio:
Implemented

PR #159, opened 2022-11-14
add certificate based verification

flavio:
This PR introduces the ability to perform signature verifications using a x509 certificate.

This fixes issue https://github.com/sigstore/sigstore-rs/issues/157

Note, some internal refactoring have been done to implement this feature. The public API of the crate has not changed.


src/cosign/verification_constraint/cert_subject_email_verifier.rs
```diff
@@ -0,0 +1,211 @@
+use super::VerificationConstraint;
```
Xynnn007, line 1: I like the refactoring here, making all the `Constraing` as submodule

src/cosign/verification_constraint/public_key_verifier.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub fn new(key_raw: &[u8], signing_scheme: &SigningScheme) -> Result<Self> {
+        let key = CosignVerificationKey::from_pem(key_raw, signing_scheme)?;
+        Ok(PublicKeyVerifier { key })
+    }
+
+    /// Create a new instance of `PublicKeyVerifier`.
+    /// The `key_raw` variable holds a PEM encoded representation of the
+    /// public key to be used at verification time. The verification
+    /// algorithm will be derived from the public key type:
+    /// * `RSA public key`: `RSA_PSS_SHA256`
```
Xynnn007, line 25: Not related to this PR. The note here is a mistake made by me. `RSA_PSS_SHA256` -> `RSA_PKCS1_SHA256`. We can fix it in another pr

src/crypto/mod.rs
```diff
@@ -206,13 +206,21 @@ OSWS1X9vPavpiQOoTTGC0xX57OojUadxF1cdQmrsiReWg2Wn4FneJfa8xw==
         pub subject_issuer: Option<String>,
         pub not_before: DateTime<chrono::Utc>,
         pub not_after: DateTime<chrono::Utc>,
+        pub private_key: pkey::PKey<pkey::Private>,
```
Xynnn007, line 209: Not related to this PR. I found that the x509 cert used here depends on openssl, although all `dev-dependency`. Do we need to convert openssl -> RustCrypto (https://docs.rs/x509-cert/latest/x509_cert/) ?

Xynnn007 (approved):

src/crypto/mod.rs
viccuad: openssl is only used in tests (hence `dev-dependency`), as we have rustls-tls selected via features for the rest. I would be glad to substitute it as dev-dependency though.

src/cosign/verification_constraint/public_key_verifier.rs
flavio: No worries, I fixed that

src/crypto/mod.rs
flavio: I'm fine moving away from the openssl crate from the unit tests, but I think this has a low priority.

We have to think about the pros and cons about this move.

From my POV:

## Pros
- Move to something that is easier to consume. The documentation and the UX of the `openssl` crate is suboptimal

## Cons
- We've already invested quite some time building wrappers around it inside of the unit test helpers
- We have access to all the features of openssl, with other libraries we might have to wait for some things to be implemented
- The certificates created inside of the unit tests use the same code a regular user would leverage when using the `openssl` cli tool. What I mean is, we can rule our potential bugs introduced by the alternative library we would pick.


The openssl crate is not user friendly and we've already invested some time building around it inside of the unit tests. We have to po

viccuad (approved):
LGTM!

src/crypto/mod.rs
viccuad: Consuming `openssl`-created certs in tests is an upside indeed, I think is worth the hassle. Let me backtrack my previous position. And given that we would need to find substitutes for all the `openssl` usages we have right now, plus it being low prio, I'm ok as it is.

PR #165, opened 2022-11-17
Set tag to COSIGN_PRIVATE_KEY_PEM_LABEL for ec.rs

danbev:

#### Summary

This commit suggests changing the tag used in `EcdsaKeys::private_key_to_encrypted_pem` from `SIGSTORE_PRIVATE_KEY_PEM_LABEL` to `COSIGN_PRIVATE_KEY_PEM_LABEL`.

The motivation for this is that using such a key with Go cosign will currently result in an `unsupported pem type` error. It seems like Go cosign is using [CosignPrivateKeyPemType](https://github.com/sigstore/cosign/blob/6b309df06f60ea5f58db22e9890713138c823d27/pkg/cosign/keys.go#L41) when [checking](https://github.com/sigstore/cosign/blob/6b309df06f60ea5f58db22e9890713138c823d27/pkg/cosign/keys.go#L208) the tag of the pem.

I'm not sure about the differences between
`SIGSTORE_PRIVATE_KEY_PEM_LABEL` and `COSIGN_PRIVATE_KEY_PEM_LABEL` and I might be missing something here but if the change in this PR is a valid then this should probably be done for rsa and ed25519 as well.

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/src/crypto/signing_key/ecdsa/ec.rs
+++ b/src/crypto/signing_key/ecdsa/ec.rs
@@ -259,7 +259,7 @@ where
                     .map_err(|e| SigstoreError::PKCS8DerError(e.to_string()))?,
             },
             _ => pem::Pem {
-                tag: SIGSTORE_PRIVATE_KEY_PEM_LABEL.to_string(),
+                tag: COSIGN_PRIVATE_KEY_PEM_LABEL.to_string(),
                 contents: kdf::encrypt(&der, password)?,
             },
         };


Xynnn007:
Hi @danbev Thanks for pointing out the bug here! It LGTM. But still some problems to ensure.

The pem label `SIGSTORE_PRIVATE_KEY_PEM_LABEL` was brought in `sigstore-rs` at the first time here https://github.com/sigstore/sigstore-rs/pull/87/files#diff-1264d08d322b15d92a0ff5f49178e377fd247b5d9e1f02f5b5cd0b928c4df6e6R66 by me.
 I referred the countepart in `sigstore-go` here https://github.com/sigstore/sigstore/blob/main/pkg/cryptoutils/privatekey.go#L41 . 

It seems that this label only shows in the definition [here](https://github.com/sigstore/sigstore/blob/main/pkg/cryptoutils/privatekey.go#L41) with name `EncryptedSigstorePrivateKeyPEMType`.

I also wonder whether this definition is useful. And I'm happy and welcome that change that `rsa`, `ed25519` together with `ec` keys all change the label to `COSIGN_PRIVATE_KEY_PEM_LABEL` and abondon the label `SIGSTORE_PRIVATE_KEY_PEM_LABEL `.

Could you give some suggestions and is there anything I ignored? 
cc @lukehinds @dekkagaijin

lukehinds:
I believe going forward we should implement the `SIGSTORE_` prefix especially with the desire to standardise client implementations. 

What are the current consequences of using `SIGSTORE_PRIVATE_KEY_PEM_LABEL`, am I right in that sigstore-rs generated keys won't work in cosign, but the reverse is true the sigstore-rs can manage cosign generated keys?

cc @znewman01 & @asraa

znewman01:
> I believe going forward we should implement the `SIGSTORE_` prefix especially with the desire to standardise client implementations.

Strong +1 to this. Filed https://github.com/sigstore/cosign/issues/2471

flavio:
> > I believe going forward we should implement the `SIGSTORE_` prefix especially with the desire to standardise client implementations.
> 
> Strong +1 to this. Filed [sigstore/cosign#2471](https://github.com/sigstore/cosign/issues/2471)

That means we don't have to do any change to our code. Do I get it right? Can this PR be closed then?

viccuad:
Agree on closing, given that there is consensus on targetting only `SIGSTORE_PRIVATE_KEY_PEM_LABEL`, and signing is a new unreleased feature; we aren't even introducing a regression on sigstore-rs.

danbev:
Thanks everyone for commenting, I'll close this now.

PR #168, opened 2022-11-24
Update path to fulcio-cert in verify example

danbev:

#### Summary

Currently, the following error will be generated when running the verify example:
```console
Cannot build sigstore repo data: Error reading fulcio certificate from disk: No such file or directory (os error 2)
```
This commit updates the path to point to fulcio.crt.pem in the same directory as rekor.pub.

#### Release Note

NONE

#### Documentation

NONE


flavio (approved):
thanks for the fix

Issue #117, opened 2022-08-31
Add examples for each of the cosign verification mechanisms


lkatalin:
The README specifies that we can:
- [x] Verify using a given key
- [ ] Verify bundle produced by transparency log (Rekor)
- [ ] Verify signature produced in keyless mode, using Fulcio Web-PKI

However, the current documented example is for verifying using a given key only. We should add two more examples.

PR #171, opened 2022-11-25
Introduce SignedArtifactBundle

danbev:
#### Summary

This commit adds a new struct named SignedArtifactBundle which represents a signed artifact bundle as produced when signing with cosign and specifying the --bundle option.

The motivation for change comes from trying to implement an example that verifies a blob using a bundle.
For example, first a blob is signed using the following command:
```console
cosign sign-blob --bundle=artifact.bundle artifact.txt
```
The `artifact.bundle` file generated by the above command will look something like this (shortened to fit the commit message format):
```json
{
  "base64Signature": "...",
  "cert": "...",
  "rekorBundle": {
    "SignedEntryTimestamp": "...",
    "Payload": {
      "body": "...",
      "integratedTime": 1669361833,
      "logIndex": 7810348,
      "logID": "..."
    }
  }
}
```

Currently, to create Bundle from this, one would have to parse the string as json, and then access the `rekorBundle` element, and then serialize it so that it can be passed to `Bundle::new_verified`.

With the changes in this commit it will be possible to call `SignedArtifactBundle::new_verified` and pass in the contents for the bundle file directly.

Refs: https://github.com/sigstore/sigstore-rs/issues/117

#### Release Note

NONE

#### Documentation

NONE


src/cosign/bundle.rs
```diff
@@ -20,23 +20,52 @@ use std::cmp::PartialEq;
 use crate::crypto::{CosignVerificationKey, Signature};
 use crate::errors::{Result, SigstoreError};
 
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub struct Bundle {
+    #[serde(rename(deserialize = "base64Signature"))]
+    pub base64_signature: String,
+    pub cert: String,
+    #[serde(rename(deserialize = "rekorBundle"))]
```
viccuad, line 28: maybe both deserialize can be generalized for the whole struct with `#[serde(rename_all = "camelCase")]`.

viccuad (approved):
Thanks for the contribution! looks good to me, modulo the serde nitpick.

src/cosign/bundle.rs
danbev: Ah good point! I'll update this shortly. Thanks!

viccuad (dismissed):

viccuad:
Indeed, this is a breaking change for `SignatureLayer{}` if there's a name change. I am also happy with `SignedArtifactBundle`.

danbev:
> I have nothing against adding this new struct, but I'm a bit reluctant to rename the current `Bundle` struct. This would break our public API, which is not a taboo, but should be pondered.
> 
> Why don't we create instead a new struct called `SignedArtifactBundle`? The bundle you're referencing is created only by the `sign-artifact` command of cosign. I think this approach would be more clear about what this new struct is about and avoid the API break.
> 
> What do you think?

Sounds good, I'll take a look at this and also add some documentation :+1:

flavio:
Thanks @danbev! :pray:

src/cosign/bundle.rs
```diff
@@ -20,6 +20,28 @@ use std::cmp::PartialEq;
 use crate::crypto::{CosignVerificationKey, Signature};
 use crate::errors::{Result, SigstoreError};
 
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct SignedArtifactBundle {
+    pub base64_signature: String,
+    pub cert: String,
```
flavio, line 27:
```suggestion
    /// PEM encoded certificate
    pub cert: String,
```

src/cosign/bundle.rs
```diff
@@ -20,6 +20,28 @@ use std::cmp::PartialEq;
 use crate::crypto::{CosignVerificationKey, Signature};
 use crate::errors::{Result, SigstoreError};
 
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
```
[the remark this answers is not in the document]
danbev: I've added some comments now. Thanks

flavio (approved):

PR #194, opened 2023-01-16
Improve public key output and add file output

### LINKED ISSUE #105: Better output for Rekor get_public_key
The [formatting](https://github.com/sigstore/sigstore-rs/blob/371b22af9d3ae49a5bcf7fd0bca963c75e39d8f5/examples/get_public_key/main.rs#L37) could be improved, and also it would be reasonable to have an option to output this to a file at a path chosen by the user.

Gronner:
#### Summary

As described in #105 the output provided by the example `get_public_key` was not pretty. It included quotation marks at the beginning and the and, as well as the newline characters `\n`. 
In addition this prevented piping the output of the example into a file, as it would include the quotation marks and the newline characters. Now using `cargo run --example get_public_key > key.pub` the public key can be stored in a file in the PEM format.

In addition it can be useful to directly store the public key in a file instead of piping to a file.  

#### Release Note

The Rekor example `get_public_key` was expanded to:
* use `Display` printing 
* provide a command line option to store the public key in a file

#### Documentation

Close #105

### DIFF
--- a/examples/rekor/get_public_key/main.rs
+++ b/examples/rekor/get_public_key/main.rs
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::{fs::write, process};
+
 use clap::{Arg, Command};
 use sigstore::rekor::apis::{configuration::Configuration, pubkey_api};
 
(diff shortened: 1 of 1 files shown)


examples/rekor/get_public_key/main.rs
```diff
@@ -38,5 +48,16 @@ async fn main() {
         flags.get_one::<String>("tree_id").map(|s| s.as_str()),
     )
     .await;
-    println!("{:#?}", pubkey.unwrap());
+
+    if let Some(out_path) = flags.get_one::<String>("output") {
+        match write(out_path, pubkey.unwrap()) {
```
flavio, line 53: nitpicking: instead of using `unwrap` use `expect` with a nice message. I would do that right on line `50`, after the `await`.

flavio (changes requested):

examples/rekor/get_public_key/main.rs
Gronner: Done, in the future it might be useful to enhance the errors get_public_key returns to enhance the possible feedback or by "unpacking" the information stored in the returned error.

flavio (approved):

PR #197, opened 2023-01-23
fix: show actual response status field

ctron:
#### Summary

Show the actual response status code, instead of `{error.status:?}`.

Currently, when a remote call fails with a client or server error, it shows a response error with the following message:

```
error in response: status code {{error.status:?}}
```

However, I do believe the actual intention (and what would be helpful), is to show the actual status code, instead of the literal string: `{error.status:?}`.

This PR fixes this issue, so that the error code gets shown.

#### Release Note

Fix an issue where, instead of the actual error code, the literal string `{error.status:?}` was shown.

### DIFF
--- a/src/rekor/apis/mod.rs
+++ b/src/rekor/apis/mod.rs
@@ -27,8 +27,13 @@ pub enum Error<T> {
         source: std::io::Error,
     },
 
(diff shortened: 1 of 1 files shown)


flavio (approved):
Thanks for the fix

PR #202, opened 2023-01-24
fix: make the fields accessible

ctron:
Signed-off-by: Jens Reimann <jreimann@redhat.com>

#### Summary

Not having access to these fields makes it hard to work with the data (like verifying).


flavio (approved):
LGTM

PR #209, opened 2023-02-03
Fulcio API v2.

### LINKED ISSUE #139: Switch to Fulcio V2 API
**Description**

The Fulcio V1 API will be turned down in the coming months.

https://github.com/sigstore/sigstore-rs/blob/main/src/fulcio/mod.rs#L20 will need to be updated to https://github.com/sigstore/fulcio/blob/main/fulcio.proto#L65. Fulcio V2 supports either gRPC or HTTP.

jvanz:
#### Summary

Updates sigstore.rs to use the latest Fulcio API v2 to request certs. This required a change in the structures used to send the request because the payload changed in the new API version.

Fix #139 

#### Release Note

Switch to the latest Fulcio API v2.


jvanz:
I still wanna test and improve the code. But, as my second contribution in the Sigstore ecosystem, I think is worth to open a draft PR to get some initial feedback from all of you. Thanks!

jvanz:
FYI: @flavio

src/fulcio/mod.rs
```diff
@@ -40,37 +84,33 @@ impl TryFrom<Csr> for Body {
     }
 }
 
-/// Internal newtype to control serde jsonification.
-#[derive(Debug)]
-struct PublicKey(String, SigningScheme);
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct Chain {
+    certificates: Vec<FulcioCert>,
```
flavio, line 90: please add the inline comment from [here](https://github.com/sigstore/fulcio/blob/1891616d6ee4f2467eda13897c8e0fb051c06c26/fulcio.proto#L201-L206)

src/fulcio/mod.rs
```diff
@@ -40,37 +84,33 @@ impl TryFrom<Csr> for Body {
...
-                | SigningScheme::RSA_PKCS1_SHA512(_) => "rsa",
-            },
-        )?;
-        pk.end()
-    }
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SignedCertificateDetachedSct {
+    chain: Chain,
+}
```
flavio, line 97: Missing inline comments and `signed_certificate_timestamp`, according to [the definition](https://github.com/sigstore/fulcio/blob/1891616d6ee4f2467eda13897c8e0fb051c06c26/fulcio.proto#L156-L173)

src/fulcio/mod.rs
```diff
@@ -40,37 +84,33 @@ impl TryFrom<Csr> for Body {
...
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SignedCertificateEmbeddedSct {
+    chain: Chain,
+}
```
flavio, line 103: Missing inline comment, please copy from [here](https://github.com/sigstore/fulcio/blob/1891616d6ee4f2467eda13897c8e0fb051c06c26/fulcio.proto#L175-L187)

src/fulcio/mod.rs
```diff
@@ -40,37 +84,33 @@ impl TryFrom<Csr> for Body {
...
+
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct CsrResponse {
+    signed_certificate_detached_sct: Option<SignedCertificateDetachedSct>,
+    signed_certificate_embedded_sct: Option<SignedCertificateEmbeddedSct>,
```
flavio, line 109: Also in this case, these attributes are mutually exclusive because protobuf defines them as `oneof` as seen [here](https://github.com/sigstore/fulcio/blob/1891616d6ee4f2467eda13897c8e0fb051c06c26/fulcio.proto#L147-L152)

flavio (commented):
> Warning: this is not related with this specific PR.

I have doubts about the Fulcio types. As I pointed out inside of the comments some of them are not mapping the protobuf specification. I think it would be nice to revisit how these types are generated and try to automate that via some dedicated crate and the usage of of a `build.rs`.

Right now I fear that keeping things in sync is just a manual process that requires quite some scrutiny to avoid this kind of mistakes

Issue #187, opened 2023-01-12
Define fine-grained feature to decrease the compiled size


Xynnn007:
**Description**

Now all the different functionalities of `sigstore` crate depends on a lot of dependecies. I suggest to divide different componants into different features. When users use the lib, they can enable different features they need to decrease the size of the final binary, rather than compile all the dependencies (some might not be used).

Xynnn007:
Sorry for reopening this issue. There should be more test and finer granularity for the dependencies... As deps which depends on `ring` can not be compiled on `s390x`, which makes trouble of the crate for downstream... Let me fix this further.

flavio:
No worries at all, thanks!

From what I can see, this is happening because `ring` is brought in as a transitive dependency of picky-rs and openid-connect.

I'm the one who caused picky-rs to use ring, I can revisit my merged code to use pure rust libraries. Back in the days there was no native ecdsa p384 crate yet. But this has now changed.

What do you think? Maybe a similar patch could be made also to openid-connect

Xynnn007:
@flavio Thanks a lot! That would be great if it is convenient! Also `x509-parser` brings `ring`. I'll try to replace `x509-parser` with a purely rust crate (I don't know if I can)

Besides, let me check further if different feature combinations could pass tests and compilation, which could cause potential problems.

Xynnn007:
As for the `openid-connect`, the newest code has abondoned `ring` with Rust Crypto. We can update the version when the stable version is released.

flavio:
> @flavio Thanks a lot! That would be great if it is convenient! Also x509-parser brings ring. I'll try to replace x509-parser with a purely rust crate (I don't know if I can)

I'll set some time aside during the next week to produce a patch for `picky-rs`

PR #211, opened 2023-02-08
fix: bring tuf feature out of rekor and add related docs

Xynnn007:
Related to https://github.com/sigstore/sigstore-rs/issues/187

#### Summary
- TUF is a independent feature thus bring out of rekor
- Added `mock-client` feature to a test which depends on `mock-client`.


flavio (approved):
LGTM

PR #212, opened 2023-02-10
Replace `x509-parser` with `x509-cert`

Xynnn007:
#### Summary

Replace `x509-parser` with `x509-cert`.

#### Release Note

- `x509-cert` is from RustCrypto, and the replaced `x509-parser` depends on `ring`. This PR will help to get rid of `ring`.
- Add a test that converts a public key cert of ed25519 into a `CosignVerificationKey`


flavio (approved):
Wow, that's super nice! Feel free to merge this PR anytime you want

Issue #214, opened 2023-02-21
Wrong parameter order inside documentation example


vembacher:
**Description**

The parameter order in the example is wrong, `source_image_digest` and `cosign_image` are switched.

https://github.com/sigstore/sigstore-rs/blob/bb68d371e8c9c4d513b7f74e2073bec80e6ffd2f/src/cosign/client.rs#L75-L80

https://github.com/sigstore/sigstore-rs/blob/bb68d371e8c9c4d513b7f74e2073bec80e6ffd2f/src/lib.rs#L103-L107

**Effect:** A `RegistryPullManifestError` as the OCI client reaches out to  `docker.io/library/<source_image_digest>`, which is hard to understand the source of if you do not closely inspect the URL in the error message.

**Version**

This affects the most recently on crates.io published version `0.6.0` and the current main branch version. Also affects earlier versions.

**Possible API improvment:** using `&oci_distribution::Reference` or a wrapper type instead of `&str` for OCI images, though I'm unsure about the consequences of this as I'm not too experienced with API design.

flavio:
> Possible API improvment: using &oci_distribution::Reference or a wrapper type instead of &str for OCI images, though I'm unsure about the consequences of this as I'm not too experienced with API design.

I think this would be a nice fix. The API would break, but this is fine.

Would you like to work on this fix with another PR?

PR #215, opened 2023-02-21
Fix: Wrong parameter order inside documentation example.

vembacher:
#### Summary

This fixes the issue described in #214 by putting the parameters into the correct order.

Resolves: #214

#### Release Note

Fixed parameter order for `sigstore::cosign::Client::trusted_signature_layers()` in documentation example causing `RegistryPullManifestError` when used.

#### Documentation

This affects the **crate documentation**, but should not affect the documentation on https://docs.sigstore.dev


flavio (approved):
thanks for the fix

PR #216, opened 2023-02-21
Introduce Newtype `OciReference` into API for OCI image references.

vembacher:
#### Summary

This introduces a Newtype `OciReference` to the public API.

There are two motivations behind this change:
- improve the API to prevent errors such as #214 through a more descriptive type.
- The Newtype is a wrapper around the internally used `oci_distribution::Reference` type in order to prevent leaking this into the public API.

See also: #214 and #215

#### Release Note

- Added type `OciReference` in `sigstore::registry::oci_reference` to represent OCI references in other APIs.
- `CosignCapabilities` trait: changed signature of functions `triangulate()`, `fetch_manifest_and_layers()`  `push_signature()`, parameters related OCI image references now have type `&OciReference` instead of `&str`.
- `SimpleSigning::new()`: changed type of `image_ref` to `&OciReference` from `&str` .
- `SignatureLayer::new_unsigned()`: changed type of parameter `image_ref` to `&OciReference` from `&str`.
- Updated examples broken by API changes both in examples folder and in the documentation.

#### Documentation

This only affects the crate docs, which I updated to address the changes to my best abilities.

Users can address the breaking API changes by via using the functions associated with the `FromStr` trait.

### DIFF
--- a/src/registry/oci_reference.rs
+++ b/src/registry/oci_reference.rs
@@ -1,0 +1,91 @@
+//
+// Copyright 2021 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::errors::SigstoreError;
+use std::fmt::{Display, Formatter};
+use std::str::FromStr;
+
+/// Reference provides a general type to represent any way of referencing images within an OCI registry.
+#[derive(Debug, Clone, PartialEq)]
+pub struct OciReference {
+    pub(crate) oci_reference: oci_distribution::Reference,
+}
+
+impl FromStr for OciReference {
+    type Err = SigstoreError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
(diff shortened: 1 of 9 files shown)


flavio:
> It also introduces a foreign type into the public API. Should the type be re-exported? Should it be wrapped via the Newtype pattern?

Ouch, I didn't realize that when I asked for the change. I would not like to expose a `oci_distribution` type inside of our public API.

I would either go with the [Newtype pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/newtype.html), as @vembacher suggests, or just drop this PR entirely.

What do the other @sigstore-rs-codeowners think?

flavio:
@vembacher sorry for the delay, if you're still interesting in moving this forward I would go with the Newtype pattern

vembacher:
@flavio I'm still interested in making this change.

I updated the code with the Newtype, there might still be some minor changes possible to where it is used. I tried to keep the changes as little as possible. I will update the summary at the top accordingly.

Let me know if you want any changes.

src/registry/oci_reference.rs
```diff
@@ -0,0 +1,91 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::errors::SigstoreError;
+use std::fmt::{Display, Formatter};
+use std::str::FromStr;
+
+/// Reference provides a general type to represent any way of referencing images within an OCI registry.
```
flavio, line 20:
```suggestion
/// `OciReference` provides a general type to represent any way of referencing images within an OCI registry.
```

flavio (approved):
That looks great to me, thanks a lot!

I left one minor suggestion

vembacher:
Thanks for the remark, I applied to change.

flavio (approved):
Thanks for your contribution

PR #221, opened 2023-03-01
Support compiling to wasm32 architectures

### LINKED ISSUE #220: Support compiling on wasm32 architectures
The ability to compile on wasm32 architectures would make it possible to use sigstore-rs in the browser, or on WASI runtimes.

lulf:
#### Summary

Enable support for compiling sigstore-rs to WASM, making it usable from browser applications and on wasm32 VMs.

Example application using the Yew framework: https://github.com/trustification/sigstore-search

Closes  #220

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,22 @@ jobs:
         with:
           command: check
 
+  check-wasm:
+    name: Check WASM
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
+        with:
+          profile: minimal
+          toolchain: stable
+          target: wasm32-unknown-unknown
+          override: true
+      - uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # v1.0.3
+        with:
+          command: check
+          args: --no-default-features --features wasm
+
   test:
     name: Test Suite
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ readme = "README.md"
 
 [features]
 default = ["full-native-tls", "cached-client", "tuf"]
+wasm = ["getrandom/js"]
 
 full-native-tls = ["fulcio-native-tls", "rekor-native-tls", "cosign-native-tls", "mock-client-native-tls"]
 full-rustls-tls = ["fulcio-rustls-tls", "rekor-rustls-tls", "cosign-rustls-tls", "mock-client-rustls-tls"]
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -55,18 +55,19 @@ cfg-if = "1.0.0"
...
 ecdsa = { version = "0.15", features = [ "pkcs8", "digest", "der" ] }
 ed25519 = { version = "2", features = [ "alloc" ] }
 ed25519-dalek = { version = "2.0.0-pre.0", features = [ "pkcs8", "rand_core" ] }
 elliptic-curve = { version = "0.12.2", features = [ "arithmetic", "pem" ] }
 lazy_static = "1.4.0"
 oci-distribution = { version = "0.9", default-features = false, optional = true }
 olpc-cjson = "0.1"
-open = "3.0.1"
 openidconnect = { version = "2.3", default-features = false, features = [ "reqwest" ], optional = true}
+getrandom = { version = "0.2.8", features = ["js"] }
```
flavio, line 67: I'm not sure this is something we have to do, quoting the documentation of the [getrandom crate](https://docs.rs/getrandom/0.2.8/getrandom/#webassembly-support):

> This feature should only be enabled for binary, test, or benchmark crates. Library crates should generally not enable this feature, leaving such a decision to users of their library. Also, libraries should not introduce their own js features just to enable getrandom’s js feature.

I think a better approach would be to:

* Extend our documentation explaining how to use this crate when one of the WebAssembly targets is used
* Move the `getrandom` dependency to the dev-dependencies. This will ensure the unit tests can be run

.github/workflows/tests.yml
```diff
@@ -17,6 +17,21 @@ jobs:
         with:
           command: check
 
+  check-wasm:
```
flavio, line 20: Should we check the `wasm32-wasi` target too?

flavio (commented):
That's cool. Would it be possible to add an example of this kind of usage?

Cargo.toml
lulf: Would it make sense to add a 'wasm' feature (and document it) to the crate that enabled it instead? I think it's a bit strange to require 'user' crates to set features on transient dependencies in order for things to compile.

.github/workflows/tests.yml
lulf: Unfortunately, the ring dependency does not work on wasm32-wasi (yet).

Cargo.toml
flavio: I've already ran into some crates that require that, so it doesn't feel so strange to me. However, I think adding a `wasm` feature would be better

lulf: I've update to enable the js flag only if the wasm feature is enabled, and also added some info to the readme.

flavio (commented):
LGTM, can be merged once all the tests are green

lulf:
Clippy and integration tests compile now, and I squashed the commits. I think the GHA should be correct as well.

flavio (approved):

Issue #222, opened 2023-03-02
Using ed25519 version = "2" leads build errror due to version conflict with d25519-dalek


vembacher:
**Description**

- This line can cause `ed25519` to be resolved to 2.2.0:
https://github.com/sigstore/sigstore-rs/blob/25dc99b3ad6709bd364d455437b88bade19c47f0/Cargo.toml#L60

- however `ed25519-dalek` requires version 2.1.0:
https://github.com/dalek-cryptography/ed25519-dalek/blob/b77fa515690db7ff0df84cea497f2f7e467adc29/Cargo.toml#L29

- because `ed25519` 2.2.0 version bumped `pkcs8` to version 0.10.0: 
https://github.com/RustCrypto/signatures/blob/b9c341a55ebf9b27ab8bdd9f10ea5b0d2cbc23a9/ed25519/Cargo.toml#L23

- which uses `spki` version 0.7.0, which changed the signature of `SubjectPublicKeyInfo`
  - https://github.com/RustCrypto/formats/blob/spki/v0.6.0/spki/src/spki.rs#L37
  - https://github.com/RustCrypto/formats/blob/spki/v0.7.0/spki/src/spki.rs#L44

For instance building with this fails for me:
```Dockerfile
FROM rust:1.67.1-buster
RUN git clone https://github.com/sigstore/sigstore-rs \
        && cd sigstore-rs \
        && cargo build
```

**Version**

Latest commit from main branch.

PR #223, opened 2023-03-02
Fix ed25519 version conflict

vembacher:
Closes: #222

#### Summary

- Refer to #222
- This fixes the issue by forcing `ed25519` version to be 2.1.0, which allows `ed25519-dalek` to be compiled against 2.1.0 instead of the incompatible 2.2.0.

#### Release Note

NONE

#### Documentation

NONE


flavio (approved):
Thanks for the fix

PR #226, opened 2023-03-08
dep: update picky version to git rid of `ring`

Xynnn007:
#### Summary

- The newest version `picky` removes `ring` dependency with RustCrypto. By this updation, we can fully get rid of `ring` dependency for cosign verification.

#### Release Note

Removed `ring` dependency when feature `["cosign-native-tls", "cached-client"]` are enabled.
#### Documentation

Close https://github.com/sigstore/sigstore-rs/issues/187

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ p256 = "0.12"
 p384 = "0.12"
 webbrowser = "0.8.4"
 pem = "1.0.2"
-picky = { version = "7.0.0-rc.3", default-features = false, features = [ "x509", "ec" ] }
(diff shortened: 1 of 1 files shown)


Xynnn007:
I'm not sure whether we can keep a git-rev in the `Cargo.toml`, or we should wait for it to be released?

flavio:
> I'm not sure whether we can keep a git-rev in the `Cargo.toml`, or we should wait for it to be released?

We have to wait for a release on crates.io. Crates that are published on crates.io (like sigstore-rs) cannot have git references.

flavio:
I'm totally in favor of this PR to be merged, let's wait until there's another RC of picky

flavio (approved):
Awesome!

PR #259, opened 2023-05-18
Address RUSTSEC-2023-0037

dmitris:
#### Summary
Address RUSTSEC-2023-0037 security advisory.
Change references of the unmaintained crate xsalsa20poly1305 to crypto_secretbox.
It should fix the failing `security_audit` check - https://github.com/sigstore/sigstore-rs/actions/runs/5014422162/jobs/8988653668.

https://rustsec.org/advisories/RUSTSEC-2023-0037.html

#### Release Note
(Bug fixes and fixes of previous known issues)
- replace xsalsa20poly1305 with crypto_secretbox to address RUSTSEC-2023-0037 security advisory

#### Documentation
- no changes

PR #260, opened 2023-05-18
update pkcs1 from 0.4.0 to 0.7.5

dmitris:
Update the [pkcs1](https://crates.io/crates/pkcs1) dependency to the latest version, `0.7.5` - with minor necessary source code adjustments.

Currently PR is cumulative with #259 - can either merge this one to cover both small changes, or I can rebase it one #259 is merged.


lukehinds (approved):
Let's merge this one, as we can see both in unison are passing CI

PR #261, opened 2023-05-19
use 2021 Rust edition

dmitris:
since this crate is new and indicated as "experimental", suggesting to use the latest 2021 edition.


lukehinds:
sounds good to me, @flavio any possible downstream impact on k-warden?

viccuad:
@lukehinds, from Kubewarden side we fine, pieces consuming sigstore-rs are already in 2021.

viccuad (approved):

PR #268, opened 2023-05-24
update scrypt to 0.11.0, adapt for API change (fix #231)

dmitris:
PR is a fixed version of #231 with the call to [scrypt::Params::new](https://docs.rs/scrypt/0.11.0/scrypt/struct.Params.html#method.new) modified for the crate's API change.


flavio (approved):
thanks for the contribution!

PR #275, opened 2023-06-01
fix: ensure cosign client can be sent between threads

flavio:
After d9b198ae4530a63932d8d674b1dd033fc08b0dfe got merged, the `Sync` and `Send` supertraits got removed from the `ClientCapabilities` trait. This was done to ensure the code could build when the `wasm32` target is used.

Unfortunately, this makes impossible to send the cosign client between threads. This causes an error like the following one:

```
error: future cannot be sent between threads safely
the trait `Send` is not implemented for `(dyn sigstore::registry::ClientCapabilities + 'static)`
```

This commit fixes the issue by introducing conditional compilation checks.

### DIFF
--- a/src/cosign/client.rs
+++ b/src/cosign/client.rs
@@ -43,7 +43,8 @@ pub struct Client {
     pub(crate) fulcio_cert_pool: Option<CertificatePool>,
 }
 
-#[async_trait(?Send)]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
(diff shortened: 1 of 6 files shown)


flavio:
Once this is merged I would like to tag the `v0.7.1` release

Xynnn007:
Hey I love this PR. Before we have to create a temporary runtime to handle this

Xynnn007 (approved):

viccuad (approved):
LGTM!

PR #285, opened 2023-07-19
Merkle tree proof implementation

vembacher:
#### Summary

This adds implementations for:

- consistency proofs
- inclusion proofs
- checkpoint/STH related functionality

The Merkle proofs are essentially ports of the [transparency-dev implementations](https://github.com/transparency-dev/merkle/tree/036047b5d2f7faf3b1ee643d391e60fe5b1defcf/proof), including the test suite.
The checkpoint related code is based on the [implementation in the `rekor` Go package](https://github.com/sigstore/rekor/tree/3bdf74677a279cc46afc1c6ef35e2afdb16058b4/pkg/util).

I also changed the Rekor models to use the new `SignedCheckpoint` type, which implements the `serde` traits. For now I think this is the only major breaking change, the rest are mostly new or private APIs.

I have not implemented the logic to verify that Checkpoints and the corresponding consistency/inclusion proof are sound together. I want to discuss how to handle this here.

#### Release Note

TODO when finalized

#### Documentation

TODO when finalized


vembacher:
I added examples for consistency and inclusion proofs.

I will also add some tests in the future to the methods I added to the Rekor models. What test data should I use for this? As these require log entries + keys.

Also I'm unsure whether [my checks to ensure whether proofs and checkpoints are valid together](https://github.com/sigstore/sigstore-rs/pull/285/commits/c1eadee306010611d7520fa45dec18621a8ac384#diff-65a1e8a6b0905f301ed949e0e47de53f91a681e8bac66197aa1fc395bb20fd25R160-R187) are correct. What is the correct way to do this?

flavio:
Sorry, this has been sitting on my review list for a long time. I need some time to get familiar with the feature being implemented :sweat:

vembacher:
> Sorry, this has been sitting on my review list for a long time. I need some time to get familiar with the feature being implemented 😓

No worries! I'm not in a rush to complete this.

gaetanww:
I received an error with the current implementation when I tested for the entry with index `25579`.

## Problem

I received the following error although the verification should have succeeded.
```
InclusionProofError(MismatchedRoot { expected: "db0822d0508b54bfb770ba5d3394d331bd9a22c91d9a8c0b906e7ea2208903be", got: "4d006aa46efcb607dd51d900b1213754c50cc9251c3405c6c2561d9d6a2f3239" })
```

## Underlying problem

The underlying error is, I believe, related to the assumption that the body is formatted as canonical JSON when it is appended to the log. Indeed, to verify that the body is included in the merkle tree, the implementation currently [reserializes the body using a canonical formatter](https://github.com/sigstore/sigstore-rs/blob/acbcb91f1e69f8aab5262fb1859a75ed9d94a43f/src/rekor/models/log_entry.rs#L131). However, I think the original entry wasn't formatted using a canonical formatter (see files below).

## To fix

We can't make the assumption that the body was formatted canonically, so I think the solution should be to keep the "original" base64-decoded body as bytes in the `LogEntry` struct and use this to verify the inclusion proof instead of reserializing it. 

If you agree with the fix, I'm happy to work on it.

## To reproduce

The following code:
```rust
let rekor_config = Configuration::default();
let rt = tokio::runtime::Runtime::new().unwrap();
let log_entry = rt.block_on(get_log_entry_by_index(&rekor_config, 25579)).unwrap();
let mut encoded = vec![];
let mut ser = serde_json::Serializer::with_formatter(
    &mut encoded,
    olpc_cjson::CanonicalFormatter::new(),
);
log_entry.body.serialize(&mut ser).unwrap();
println!("{}", String::from_utf8(encoded).unwrap());
```
returns:
```json
{"apiVersion":"0.0.1","kind":"rekord","spec":{"data":{"hash":{"algorithm":"sha256","value":"ce9a7c82f32194995888758cf107ef0cc52e0b8cdce73b4240658ee9e73783cb"}},"signature":{"content":"MGUCMD3oKzgsGnPAkJEXegDIsdlh4BFCQbM6jng4Sy3axY/+2tlK97oe/CkxabT1ZXUqCAIxAJDq+zLfRZZEJD5DvaKhFEu+Jm+jD4UXc3CaZp2MSajiralmtalA6fSGCXjwGfUzOw==","format":"x509","publicKey":{"content":"LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUNrakNDQWhpZ0F3SUJBZ0lVQU0rK0dYRFN5bUNPSW82YmxMMG5EZngxb21nd0NnWUlLb1pJemowRUF3TXcKS2pFVk1CTUdBMVVFQ2hNTWMybG5jM1J2Y21VdVpHVjJNUkV3RHdZRFZRUURFd2h6YVdkemRHOXlaVEFlRncweQpNVEEzTWpnd09ETTNOREphRncweU1UQTNNamd3T0RVM05ERmFNQUF3ZGpBUUJnY3Foa2pPUFFJQkJnVXJnUVFBCklnTmlBQVJjMDMrUU4vTHBrOGpqUFQwTmV5a01ucm9mMnpZUkJxNm05ei9TMXhRSzduZnZhU3grRjUrTEtwN3gKR2ExbHY2SWNvRXdwUHA2MUdsYnd5S0VQVWJLdzJrYnJyRVpPMnhKV3kxb0VEUHBYMlJqcTBYS0RZcEF5Zi9mQwoyZzJjSnVtamdnRW5NSUlCSXpBT0JnTlZIUThCQWY4RUJBTUNCNEF3RXdZRFZSMGxCQXd3Q2dZSUt3WUJCUVVICkF3TXdEQVlEVlIwVEFRSC9CQUl3QURBZEJnTlZIUTRFRmdRVTRBUDhtTkI4ejhSZFJyTVVLZ1A2Mm0xUFErd3cKSHdZRFZSMGpCQmd3Rm9BVXlNVWRBRUdhSkNreVVTVHJEYTVLN1VvRzArd3dnWTBHQ0NzR0FRVUZCd0VCQklHQQpNSDR3ZkFZSUt3WUJCUVVITUFLR2NHaDBkSEE2THk5d2NtbDJZWFJsWTJFdFkyOXVkR1Z1ZEMwMk1ETm1aVGRsCk55MHdNREF3TFRJeU1qY3RZbVkzTlMxbU5HWTFaVGd3WkRJNU5UUXVjM1J2Y21GblpTNW5iMjluYkdWaGNHbHoKTG1OdmJTOWpZVE0yWVRGbE9UWXlOREppT1daallqRTBOaTlqWVM1amNuUXdIZ1lEVlIwUkFRSC9CQlF3RW9FUQpZM1JoWkdWMVFHZHRZV2xzTG1OdmJUQUtCZ2dxaGtqT1BRUURBd05vQURCbEFqRUE3TTJwSzhRUFRrSGs1bzZ0CmdnampZdjBLV1BUajRKUTAwU3RjR0xqa1g3SU1iNC9HdXpYRkQ4czZDOEd3NmpwMEFqQW1Xa2JROTVsMzlnUGQKR2pjUjBSQURaT0dYb0NPQURwOE5lSzhBL2dKdWdnR0ZINHZYZ2l1ODJsQm5MOEZSc09jPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg=="}}}}
```

when a direct call to the api:
```bash
curl 'https://rekor.sigstore.dev/api/v1/log/entries?logIndex=25579' | jq -r '.[ (keys_unsorted)[0] ].body' | base64 -d
```
returns:
```json
{"apiVersion":"0.0.1","spec":{"data":{"hash":{"algorithm":"sha256","value":"ce9a7c82f32194995888758cf107ef0cc52e0b8cdce73b4240658ee9e73783cb"}},"signature":{"content":"MGUCMD3oKzgsGnPAkJEXegDIsdlh4BFCQbM6jng4Sy3axY/+2tlK97oe/CkxabT1ZXUqCAIxAJDq+zLfRZZEJD5DvaKhFEu+Jm+jD4UXc3CaZp2MSajiralmtalA6fSGCXjwGfUzOw==","format":"x509","publicKey":{"content":"LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUNrakNDQWhpZ0F3SUJBZ0lVQU0rK0dYRFN5bUNPSW82YmxMMG5EZngxb21nd0NnWUlLb1pJemowRUF3TXcKS2pFVk1CTUdBMVVFQ2hNTWMybG5jM1J2Y21VdVpHVjJNUkV3RHdZRFZRUURFd2h6YVdkemRHOXlaVEFlRncweQpNVEEzTWpnd09ETTNOREphRncweU1UQTNNamd3T0RVM05ERmFNQUF3ZGpBUUJnY3Foa2pPUFFJQkJnVXJnUVFBCklnTmlBQVJjMDMrUU4vTHBrOGpqUFQwTmV5a01ucm9mMnpZUkJxNm05ei9TMXhRSzduZnZhU3grRjUrTEtwN3gKR2ExbHY2SWNvRXdwUHA2MUdsYnd5S0VQVWJLdzJrYnJyRVpPMnhKV3kxb0VEUHBYMlJqcTBYS0RZcEF5Zi9mQwoyZzJjSnVtamdnRW5NSUlCSXpBT0JnTlZIUThCQWY4RUJBTUNCNEF3RXdZRFZSMGxCQXd3Q2dZSUt3WUJCUVVICkF3TXdEQVlEVlIwVEFRSC9CQUl3QURBZEJnTlZIUTRFRmdRVTRBUDhtTkI4ejhSZFJyTVVLZ1A2Mm0xUFErd3cKSHdZRFZSMGpCQmd3Rm9BVXlNVWRBRUdhSkNreVVTVHJEYTVLN1VvRzArd3dnWTBHQ0NzR0FRVUZCd0VCQklHQQpNSDR3ZkFZSUt3WUJCUVVITUFLR2NHaDBkSEE2THk5d2NtbDJZWFJsWTJFdFkyOXVkR1Z1ZEMwMk1ETm1aVGRsCk55MHdNREF3TFRJeU1qY3RZbVkzTlMxbU5HWTFaVGd3WkRJNU5UUXVjM1J2Y21GblpTNW5iMjluYkdWaGNHbHoKTG1OdmJTOWpZVE0yWVRGbE9UWXlOREppT1daallqRTBOaTlqWVM1amNuUXdIZ1lEVlIwUkFRSC9CQlF3RW9FUQpZM1JoWkdWMVFHZHRZV2xzTG1OdmJUQUtCZ2dxaGtqT1BRUURBd05vQURCbEFqRUE3TTJwSzhRUFRrSGs1bzZ0CmdnampZdjBLV1BUajRKUTAwU3RjR0xqa1g3SU1iNC9HdXpYRkQ4czZDOEd3NmpwMEFqQW1Xa2JROTVsMzlnUGQKR2pjUjBSQURaT0dYb0NPQURwOE5lSzhBL2dKdWdnR0ZINHZYZ2l1ODJsQm5MOEZSc09jPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg=="}}},"kind":"rekord"}
```

Note that the the `"kind"` field is at the end of the json in the API and in second place in the canonicalized format.

gaetanww:
np! Thanks for the pointer, yes that would do it! I think the solution I suggested would break backward compatibility. I think the following code should work (using the `json_syntax` crate with features `["canonicalize", "serde_json"]`)

```rust
pub fn verify_inclusion(&self, rekor_key: &CosignVerificationKey) -> Result<(), SigstoreError> {
        self.verification
            .inclusion_proof
            .as_ref()
            .ok_or(UnexpectedError("missing inclusion proof".to_string()))
            .and_then(|proof| {
                // encode as canonical JSON
                let mut json_value = json_syntax::Value::from_serde_json(serde_json::to_value(&self.body)?);
                json_value.canonicalize();
                let encoded_entry = json_value.compact_print().to_string().into_bytes();
                proof.verify(&encoded_entry, rekor_key)
            })
    }
```

vembacher:
I hopefully should be able to take a closer look at this some time next week, and apply the fix/update the PR.

gaetanww:
One more data point on issue above. It works perfectly without modification with index: `71844338`.

vembacher:
I haven't looked at it in a few months, but if I remember correctly there are two types of indices. With one being the index in the log and one being the index in the tree. So you might be using the incorrect index.

vembacher:
@gaetanww is your issue resolved?

gaetanww:
Sorry for the slow response. I made it work for the latest rekor records (e.g., index `71844338`), the early one (`25579`) still fails AFAICT.

vembacher:
I did some investigating, I think @gaetanww your original analysis was correct and the formatting of the log entry is the issue.

The [Go client also uses the Base64 encoded entry provided by the log](https://github.com/sigstore/rekor/blob/6cdd40d56e924cf13e09b40d53a04faed5e3e8c6/pkg/verify/verify.go#L158), is this the correct way to handle this? I feel like this might conflict with [the description on how SETs are supposed to be verified](https://github.com/sigstore/rekor/blob/6cdd40d56e924cf13e09b40d53a04faed5e3e8c6/openapi.yaml#L479-L482).

gaetanww:
Yes, the documentation does conflict with the actual definition. There should at least be a comment in the go code to explain why it's implemented that way. 
I think we should raise the point with the maintainers of the cosign project. I'll open an issue.

exFalso:
This is failing on 32bit targets because of the use of `usize` for bitwise arithmetic. Changing to `u64` fixes it.

tannaurus:
Howdy 👋 is there someone in the community equipped to adequately review this MR? We at 1Password would be happy to review elements of this work but the core team that has directed energy towards this project doesn't have previous experience with merkle tree proof implementations.

loosebazooka:
@tannaurus I'm going to look at the merkle tree proof as I previously did a lot of work on a different client there. However I would need someone else to do the rust review.

vembacher:
> This is failing on 32bit targets because of the use of `usize` for bitwise arithmetic. Changing to `u64` fixes it.

I will update it to `u64`, it seems like [RFC6962](https://datatracker.ietf.org/doc/html/rfc6962) also prescribes a `uint64`.

loosebazooka:
Sorry my ability to read rust is bad.

I'm a little confused by the edits in rekor/models -- those seem like generated files? But the added code doesn't seem generated? Maybe I'm just reading rust wrong.

The inclusion proof looks like its following the spec in 6962, there's a slightly different implementation in [9162](https://www.rfc-editor.org/rfc/rfc9162.html#name-merkle-inclusion-proofs) -- I don't know if it matters but it's not recursive.

It might be valuable to add some tests on the rekor inclusion proof itself (rather than the sub-functions):

one random example (from the java client):
```
      "inclusionProof": {
        "hashes": [
          "810320ec3029914695826d60133c67021f66ee0cfb09a6f79eb267ed9f55de2c",
          "67e9d9f66f0ad388f7e1a20991e9a2ae3efad5cbf281e8b3d2aaf1ef99a4618c",
          "16a106400c53465f6e18c2475df6ba889ca30f5667bacf32b1a5661f14a5080c",
          "b4439e8d71edbc96271723cb7a969dd725e23e73d139361864a62ed76ce8dc11",
          "49b3e90806c7b63b5a86f5748e3ecb7d264ea0828eb74a45bc1a2cd7962408e8",
          "5059ad9b48fa50bd9adcbff0dd81c5a0dcb60f37e0716e723a33805a464f72f8",
          "6c2ce64219799e61d72996884eee9e19fb906e4d7fa04b71625fde4108f21762",
          "784f79c817abb78db3ae99b6c1ede640470bf4bb678673a05bf3a6b50aaaddd6",
          "c6d92ebf4e10cdba500ca410166cd0a8d8b312154d2f45bc4292d63dea6112f6",
          "1768732027401f6718b0df7769e2803127cfc099eb130a8ed7d913218f6a65f6",
          "0da021f68571b65e49e926e4c69024de3ac248a1319d254bc51a85a657b93c33",
          "bc8cf0c8497d5c24841de0c9bef598ec99bbd59d9538d58568340646fe289e9a",
          "be328fa737b8fa9461850b8034250f237ff5b0b590b9468e6223968df294872b",
          "6f06f4025d0346f04830352b23f65c8cd9e3ce4b8cb899877c35282521ddaf85"
        ],
        "logIndex": 1227,
        "rootHash": "effa4fa4575f72829016a64e584441203de533212f9470d63a56d1992e73465d",
        "treeSize": 14358,
        "checkpoint": "rekor.sigstage.dev - 108574341321668964\n14358\n7/pPpFdfcoKQFqZOWERBID3lMyEvlHDWOlbRmS5zRl0=\n\n— rekor.sigstage.dev 0y8wozBFAiB8OkuzdwlL6/rDEu2CsIfqmesaH/KLfmIMvlH3YTdIYgIhAPFZeXK6+b0vbWy4GSU/YZxiTpFrrzjsVOShN4LlPdZb\n"
      },
```

and then another failure test where you mangle this data a bit.

vembacher:
@loosebazooka 
> I'm a little confused by the edits in rekor/models -- those seem like generated files? But the added code doesn't seem generated? Maybe I'm just reading rust wrong.

Yeah afaik it's a mix of auto-generated files from the OpenAPI spec which was then modified by hand.
 
> The inclusion proof looks like its following the spec in 6962, there's a slightly different implementation in [9162](https://www.rfc-editor.org/rfc/rfc9162.html#name-merkle-inclusion-proofs) -- I don't know if it matters but it's not recursive.

If my understanding is correct the semantics of the Merkle trees between 6962 and 9162 are identical. Also the [Go cosign implementation uses an implementation based on RFC 6962](https://github.com/sigstore/cosign/blob/780780b11e0998512c034317fd7e98776153e59d/go.mod#L49).

> It might be valuable to add some tests on the rekor inclusion proof itself (rather than the sub-functions):
> 
> one random example (from the java client):
> 
> and then another failure test where you mangle this data a bit.

Good point! I added more tests based on Rekor responses to the code.

vembacher:
The inclusion proofs seem to work as expected, but I'm having issues with the consistency proofs. The doctest example fails as of right now. I'm pretty sure it used to work in the past, but I'm not 100% certain. I added a test that to my understanding should be accepted, but it fails.

Did something on the Rekor side change?

vembacher:
I tracked down the issue. The following sequence of events triggers it:

1. Log info is requested, response has `root_hash_1`.
2. Something is entered to the log, leading to a new `root_hash_2`.
3. Consistency proof for tree with `root_hash_1` is requested, response has `root_hash_2` (as it is the root hash of the tree at the time the proof was requested, this is the specified behavior).
4. The proof implementation fails as it incorrectly uses `root_hash_2` instead of using `root_hash_1`.

This is a just bug in my implementation.

vembacher:
I pushed a fix that should address the issue. I still want to improve the tests for the consistency proofs.

tannaurus:
I know there's been _so_ much work here. What is holding this PR back right now?

vembacher:
@tannaurus I'm not really sure tbh.

The project for which I implemented this has completed, so I don't really have the time anymore to do the heavy lifting to get this PR done by myself without assistance from maintainers. Because I'd be doing it in my free time.

But if someone gives me a list of tasks to complete to get this merged I'm still willing to spend some time to get this done.

src/crypto/merkle/proof_verification.rs
```diff
@@ -0,0 +1,862 @@
...
+            ret.push(ConsistencyProbe {
+                size1,
+                size2,
+                root1,
+                root2,
+                proof: proof[..ln - 1].to_vec(),
+                desc: "truncated proof",
+            });
+        }
+        // add probes with proves that have a flipped 4th bit of i-th byte of the i-th hash
```
rozbb, line 724: *3rd bit

src/rekor/models/inclusion_proof.rs
```diff
@@ -25,6 +32,7 @@ pub struct InclusionProof {
     /// A list of hashes required to compute the inclusion proof, sorted in order from leaf to root
     #[serde(rename = "hashes")]
     pub hashes: Vec<String>,
```
rozbb, line 34: Same point here re storing hex vs actual bytes

rozbb (commented):
Left some very small notes. Looks great!

src/rekor/models/log_info.rs
```diff
@@ -39,4 +44,235 @@ impl LogInfo {
...
+                "treeSize": 31667593
+            }
+            ],
+            "rootHash": "e222aa53db49893334fb5a878ead1bf8b9f8f3c02ccfc0ae687f28256bd74907",
+            "signedTreeHead": "rekor.sigstage.dev - 8202293616175992157\n1352760\n4iKqU9tJiTM0+1qHjq0b+Ln488Asz8CuaH8oJWvXSQc=\n\n— rekor.sigstage.dev 0y8wozBFAiEAnIjdHAH9uhqBrRNBA4bMaKR30H6qdzW4TAsdB0/KP0ICIDjK9VeE+9dWXSAm/B0aPkhO7pJMLmKPjo9btFD9ZvEs\n",
+            "treeID": "8202293616175992157",
+            "treeSize": 1352760
+        }"#;
+
+    const LOG_INFO_NEW: &str = r#"
```
tannaurus, line 151: In an effort to get movement here I've opened a [PR in the transparency-dev/merkle](https://github.com/transparency-dev/merkle/pull/171) repo that writes their test data into json files. We could use those files in this PR to gain some confidence on the solution here. What do you think @vembacher?

tannaurus: If we do that, combined with @rozbb 's review, it feels like this PR is in a great state. Is there anything else you'd want to see done @flavio ?

wolfv:
@vembacher I merged main into your branch, resolved merge conflicts and added tests to run against @tannaurus upstream suite. Tests are passing in #527 🎉 

The main branch is currently using `olpc-json` for the serializer. Tests seem to be working with it and if I understand the above comments correctly, then the bug with was not actually in the serialization in sigstore-rs but in the formatting of the log entry? So probably, using olpc-cjson is fine?

I am waiting for @viccuad to let me know whether he would like me to look into @rozbb's comments as well. 

I would love to get this merged so that we can continue to work towards v0.3 bundle support!

src/rekor/models/checkpoint.rs
```diff
@@ -0,0 +1,445 @@
...
+    }
+}
+
+impl CheckpointSignature {
+    fn encode(&self) -> String {
+        let sig_b64 =
+            BASE64_STANDARD.encode([self.key_fingerprint.as_slice(), self.raw.as_slice()].concat());
+        format!("— {} {sig_b64}\n", self.name)
+    }
+    fn decode(s: &str) -> Result<Self, ParseCheckpointError> {
```
flavio: nit: I would add an empty line

Not a nit: add some documentation explaining what is the expected format. We could copy paste https://github.com/sigstore/rekor/blob/4b1fa6661cc6dfbc844b4c6ed9b1f44e7c5ae1c0/pkg/util/signed_note.go#L129-L140 and add a reference to it (the actual URL I just pasted).

src/rekor/models/checkpoint.rs
```diff
@@ -0,0 +1,445 @@
...
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt::Write;
+use std::fmt::{Display, Formatter};
+
+/// A checkpoint (also known as a signed tree head) that served by the log.
+/// It represents the log state at a point in time.
+/// The `note` field stores this data,
+/// and its authenticity can be verified with the data in `signature`.
+#[derive(Debug, PartialEq, Clone, Eq)]
+pub struct Checkpoint {
```
flavio: This is what the Go implementation calls [`SignedCheckpoint`](https://github.com/sigstore/rekor/blob/4b1fa6661cc6dfbc844b4c6ed9b1f44e7c5ae1c0/pkg/util/checkpoint.go#L106C1-L106C31).

This is evident when looking at the code that unmarshalls the [go SignedNote](https://github.com/sigstore/rekor/blob/4b1fa6661cc6dfbc844b4c6ed9b1f44e7c5ae1c0/pkg/util/signed_note.go#L141) (which is embedded into the Go `SignedCheckpoint`); that's the same logic used by the Rust [`Checkpoint.decode`](https://github.com/sigstore/sigstore-rs/blob/147cf7d1b1985a79ccdabbeed19dadf43f4fe423/src/rekor/models/checkpoint.rs#L71).

I find this discrepancy between the Go types and the Rust types super confusing. It makes comparing the Go implementation with ours harder.

I would stick with the Go naming

src/rekor/models/checkpoint.rs
```diff
@@ -0,0 +1,445 @@
...
+        let other_content: String = self.other_content.iter().fold(String::new(), |mut acc, c| {
+            writeln!(acc, "{c}").expect("failed to write to string");
+            acc
+        });
+        format!(
+            "{}\n{}\n{hash_b64}\n{other_content}",
+            self.origin, self.size
+        )
+    }
+    fn unmarshal(s: &str) -> Result<Self, ParseCheckpointError> {
```
flavio: I would copy-paste [this documentation](https://github.com/sigstore/rekor/blob/4b1fa6661cc6dfbc844b4c6ed9b1f44e7c5ae1c0/pkg/util/checkpoint.go#L61-L72) and add that link too.

src/rekor/models/checkpoint.rs
```diff
@@ -0,0 +1,445 @@
...
+        // refer to: https://github.com/sigstore/rekor/blob/d702f84e6b8b127662c5e717ee550de1242a6aec/pkg/util/checkpoint.go
```
flavio: nit, this link is too generic, you still have to go spelunking into the file to find what we're talking about. I would remove it (see previous comment)

src/rekor/models/checkpoint.rs
```diff
@@ -0,0 +1,445 @@
...
+        let s = s.trim_start_matches('\n').trim_end_matches('\n');
+        let [_, name, sig_b64] = s.split(' ').collect::<Vec<_>>()[..] else {
+            return Err(DecodeError(format!("unexpected signature format {s:?}")));
+        };
+        let sig = BASE64_STANDARD
+            .decode(sig_b64.trim_end())
+            .map_err(|_| DecodeError("failed to decode signature".to_string()))?;
+
+        // first four bytes of signature are fingerprint of key
+        let (key_fingerprint, sig) = sig.split_at(4);
```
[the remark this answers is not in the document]
flavio: Note `split_at` is also used elsewhere, please replace it with `split_at_checked`

src/rekor/models/checkpoint.rs
viccuad: Fixed in https://github.com/sigstore/sigstore-rs/pull/285/commits/bc19a18a62a2d85f35bb882dd252cb54e89f03a0.

src/rekor/models/checkpoint.rs
```diff
@@ -0,0 +1,445 @@
...
+        let signature = CheckpointSignature::decode(signature)?;
+        let note = CheckpointNote::unmarshal(note)?;
+
+        Ok(Checkpoint { note, signature })
+    }
+
+    pub(crate) fn encode(&self) -> String {
+        let note = self.note.marshal();
+        let signature = self.signature.encode();
+        format!("{note}\n{signature}")
```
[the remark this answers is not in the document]
viccuad: Fixed in https://github.com/sigstore/sigstore-rs/pull/285/commits/abfefddaf5726a1fd39e1182e9928cf02940c30f.

src/rekor/models/checkpoint.rs
```diff
@@ -0,0 +1,445 @@
...
+    pub(crate) fn decode(s: &str) -> Result<Self, ParseCheckpointError> {
+        // refer to: https://github.com/sigstore/rekor/blob/d702f84e6b8b127662c5e717ee550de1242a6aec/pkg/util/checkpoint.go
+
+        let checkpoint = s.trim_start_matches('"').trim_end_matches('"');
+
+        let Some((note, signature)) = checkpoint.split_once("\n\n") else {
+            return Err(DecodeError("unexpected checkpoint format".to_string()));
+        };
+
+        let signature = CheckpointSignature::decode(signature)?;
```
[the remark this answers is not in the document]
viccuad: Fixed in https://github.com/sigstore/sigstore-rs/pull/285/commits/771872c2c0e9ab5fd19393c823605b7122d20e39.

src/rekor/models/checkpoint.rs
```diff
@@ -0,0 +1,445 @@
...
+        format!(
+            "{}\n{}\n{hash_b64}\n{other_content}",
+            self.origin, self.size
+        )
+    }
+    fn unmarshal(s: &str) -> Result<Self, ParseCheckpointError> {
+        // refer to: https://github.com/sigstore/rekor/blob/d702f84e6b8b127662c5e717ee550de1242a6aec/pkg/util/checkpoint.go
+        // note is separated by new lines
+        let split_note = s.split('\n').collect::<Vec<_>>();
+        let [origin, size, hash_b64, other_content @ ..] = split_note.as_slice() else {
```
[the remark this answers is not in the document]
viccuad: Done in https://github.com/sigstore/sigstore-rs/pull/285/commits/b6af9ca5ff9075aee3b481fd9d2e560a73313d28.

vembacher:
Hello everyone, thanks for keeping this PR alive!

> @vembacher: would you be ok with trying a rebase and correct DCO of this PR?
If not, I will take care of it by just fixing the conflicts, but potentially by opening a new PR with your changes already rebased.

Can I still help with this? I tried to run the rebase+signoff command but it made go through a lot of conflicts. I haven't kept up with the development of the crate so I did not feel confident to merge these correctly and from my understanding I would have to merge them manually. Is this correct?

I tried to get an overview of the current state of the PR, but I'm not sure if I missed something. But let me know if there is anything else I can assist you with in this PR or in the corresponding [PR for the fork](https://github.com/vembacher/sigstore-rs/pull/1).

I will try my best to be assistive, but I unfortunately only have limited time to spend on this. But I will do my best to be responsive.

Ping: @viccuad @flavio @wolfv

viccuad:
I have refactored to use bytestrings for hashes, move their decoding and checks to the API boundary, and also renamed the checkpoint structs to match the Go implementation. These were the last 2 missing topics.

I consider this ready for a final review!

PR #286, opened 2023-07-26
chore(deps): do not enable default features of chrono

flavio:
Enabling the default features of chrono causes the time 0.1 crate to be added as a transitive dependency.

This old version of time is affected by CVE RUSTSEC-2020-0071

Thanks to work done inside of chrono 0.4, there are high chances the majority of the codebases do not actually need it.

Building sigstore with only the cosign feature prevents the inclusion of the vulnerable time dependency.

This isn't unfortunately true when rekor is being used, because the openid crate brings the transitive dependency back.


viccuad (approved):
LGTM! thanks

Issue #299, opened 2023-09-18
Allow for Regex in CertSubjectEmailVerifier


dave-tucker:
**Description**

In the `cosign` binary you can use `--certificate-identity-regexp` and `--certifcate-oidc-issuer-regex` to provide a regular expression for identity (email) or issuer. It would be great if we can do this from sigstore-rs also.

PR #300, opened 2023-09-18
cosign: Allow use of regex in CertSubjectEmailVerifier

dave-tucker:
#### Summary

This allows for either an exact match [StringVerifier::ExactMatch]
or it allows for a regular expression [StringVerifier::Regex]

This supports the use case of trusting signatures from a
collection of email addresses e.g .*@redhat.com and or from a
collection of issuers.

Fixes: #299

#### Release Note

- `CertSubjectEmailVerifier` is now constructed using either a `StringVerifer` enum. This supports exact string matches or regular expressions.

#### Documentation

Docstrings and examples are updated to show how this is used.

### DIFF
--- a/src/cosign/verification_constraint/cert_subject_email_verifier.rs
+++ b/src/cosign/verification_constraint/cert_subject_email_verifier.rs
@@ -1,3 +1,6 @@
+use regex::Regex;
+use std::fmt::Debug;
+
 use super::VerificationConstraint;
 use crate::cosign::signature_layers::{CertificateSubject, SignatureLayer};
 use crate::errors::Result;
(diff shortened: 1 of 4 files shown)


flavio (commented):
I like the idea, I'm in favor of implementing this change.

The code looks good, but I have some suggestions. These might be a matter of personal taste, hence I'm open to have a discussion about how to proceed.

My proposal would be to change the `StringVerifier` from being a trait to be a simple enum. We could have something like `StringVerifier::ExactMatch(String)` and `StringVerifier::Regex(Regex)`. This would make the code easier to understand for our end consumers and we could get rid of the dynamic dispatch introduced by the usage of the trait.

What do you think?

dave-tucker:
> I like the idea, I'm in favor of implementing this change.
> 
> The code looks good, but I have some suggestions. These might be a matter of personal taste, hence I'm open to have a discussion about how to proceed.
> 
> My proposal would be to change the `StringVerifier` from being a trait to be a simple enum. We could have something like `StringVerifier::ExactMatch(String)` and `StringVerifier::Regex(Regex)`. This would make the code easier to understand for our end consumers and we could get rid of the dynamic dispatch introduced by the usage of the trait.
> 
> What do you think?

Makes sense to me. I've updated the patch with your proposal.

src/cosign/verification_constraint/cert_subject_email_verifier.rs
```diff
@@ -91,23 +111,69 @@ use crate::errors::Result;
...
+            &self.email.to_string(),
+            issuer_str
+        ))
+    }
+}
+
+pub enum StringVerifier {
+    ExactMatch(String),
+    Regex(Regex),
+}
```
flavio, line 136: Can you add some inline docs to be included on docs.rs please?

src/cosign/verification_constraint/cert_subject_email_verifier.rs
```diff
@@ -203,9 +269,113 @@ mod tests {
...
         let vc = CertSubjectEmailVerifier {
-            email: "alice@example.com".to_string(),
+            email: StringVerifier::ExactMatch("alice@example.com".to_string()),
+            issuer: None,
+        };
+        assert!(!vc.verify(&sl).unwrap());
+    }
+
+    #[test]
+    fn cert_email_verifier_only_email_regex() {
```
flavio, line 279: nitpicking, using something like `rstest` could reduce a lot of this repetitions

src/cosign/verification_constraint/cert_subject_email_verifier.rs
```diff
@@ -203,9 +269,113 @@ mod tests {
...
+
+        let vc = CertSubjectEmailVerifier {
+            email: StringVerifier::ExactMatch("different@email.com".to_string()),
             issuer: None,
         };
         assert!(!vc.verify(&sl).unwrap());
     }
+
+    #[test]
+    fn cert_email_verifier_email_and_issuer_regex() {
```
flavio, line 309: same here

flavio (changes requested):
I like it, thanks for having done the proposed change.

Aside from the nitpick comment (which would be great to see addressed, but I don't consider that mandatory), can you please address the inline docs feedback please?

flavio (approved):
I realized this PR got stuck. I solved the conflicts with the `main` branch.

I'll merge it once all the tests are green

src/cosign/verification_constraint/cert_subject_email_verifier.rs
flavio: I was wrong, it's not easy to blend the two tests with rstest

src/cosign/verification_constraint/cert_subject_email_verifier.rs
flavio: I was wrong, it's not easy to blend the two tests with rstest

src/cosign/verification_constraint/cert_subject_email_verifier.rs
flavio: Looking back, that's not really needed

flavio:
@Xynnn007 can you please give me an approval? I had to fix the conflicts with the `main` branch, but I didn't do any change to the original code submitted by @dave-tucker

Xynnn007 (approved):

flavio:
@Xynnn007 thanks!

@dave-tucker sorry about the delay!

PR #311, opened 2023-11-16
verify: init

jleightcap:
#### Summary

Complementary half of #310, enable the verification of bundles:

```sh
$ cargo run --manifest-path tests/conformance/Cargo.toml -- sign-bundle --bundle bundle.txt.sigstore --identity-token $(sigstore-python get-identity-token) bundle.txt
# ...
Operation succeeded!
$ cargo run --manifest-path tests/conformance/Cargo.toml -- verify-bundle --bundle bundle.txt.sigstore --certificate-identity jack@leightcap.com --certificate-oidc-issuer https://github.com/login/oauth bundle.txt
# ...
Operation succeeded!
```

### DIFF
--- a/src/trust/sigstore/mod.rs
+++ b/src/trust/sigstore/mod.rs
@@ -22,7 +22,7 @@
 //! to enable Fulcio and Rekor integrations.
 use futures_util::TryStreamExt;
 use sha2::{Digest, Sha256};
-use std::path::PathBuf;
+use std::path::Path;
 use tokio_util::bytes::BytesMut;
 
 use sigstore_protobuf_specs::dev::sigstore::{
@@ -48,10 +48,10 @@ impl SigstoreTrustRoot {
     /// Constructs a new trust root from a [`tough::Repository`].
     async fn from_tough(
         repository: &tough::Repository,
-        checkout_dir: Option<PathBuf>,
+        checkout_dir: Option<&Path>,
     ) -> Result<Self> {
         let trusted_root = {
-            let data = Self::fetch_target(repository, &checkout_dir, "trusted_root.json").await?;
+            let data = Self::fetch_target(repository, checkout_dir, "trusted_root.json").await?;
             serde_json::from_slice(&data[..])?
         };
 
@@ -59,7 +59,7 @@ impl SigstoreTrustRoot {
     }
 
     /// Constructs a new trust root backed by the Sigstore Public Good Instance.
-    pub async fn new(cache_dir: Option<PathBuf>) -> Result<Self> {
+    pub async fn new(cache_dir: Option<&Path>) -> Result<Self> {
         // These are statically defined and should always parse correctly.
         let metadata_base = url::Url::parse(constants::SIGSTORE_METADATA_BASE)?;
         let target_base = url::Url::parse(constants::SIGSTORE_TARGET_BASE)?;
@@ -79,7 +79,7 @@ impl SigstoreTrustRoot {
 
     async fn fetch_target<N>(
         repository: &tough::Repository,
-        checkout_dir: &Option<PathBuf>,
+        checkout_dir: Option<&Path>,
         name: N,
     ) -> Result<Vec<u8>>
     where
@@ -238,6 +238,7 @@ fn is_timerange_valid(range: Option<&TimeRange>, allow_expired: bool) -> bool {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use rstest::{fixture, rstest};
     use std::fs;
     use std::path::Path;
     use std::time::SystemTime;
@@ -265,70 +266,41 @@ mod tests {
         );
     }
 
-    macro_rules! impl_test {
-        ($name:ident, cache=$cache:literal $(,setup=$setup:expr)? $(,verify=$verify:expr)?) => {
-            #[test]
-            fn $name() -> Result<()> {
-                let rt = tokio::runtime::Builder::new_current_thread()
-                    .enable_all()
-                    .build()?;
-
-                let cache_dir = if $cache {
-                    let tmp = TempDir::new().expect("cannot create temp cache dir");
-                    Some(tmp.into_path())
-                } else {
-                    None
-                };
-                let cache_dir_borrowed = cache_dir.as_ref().map(|p| p.as_path());
-
-                $(
-                    $setup(cache_dir_borrowed);
-                )?
-
-                let root = rt
-                    .block_on(SigstoreTrustRoot::new(cache_dir.clone()))
-                    .expect("failed to create trust root");
-
-                verify(&root, cache_dir_borrowed);
-
-                $($verify(&root, cache_dir_borrowed);)?
-
-                Ok(())
-            }
-        }
+    #[fixture]
+    fn cache_dir() -> TempDir {
+        TempDir::new().expect("cannot create temp cache dir")
     }
 
-    impl_test!(no_local_cache, cache = false);
-    impl_test!(
-        current_local_cache,
-        cache = true,
-        verify = |_, cache_dir: Option<&Path>| {
-            assert!(
-                cache_dir.unwrap().join("trusted_root.json").exists(),
-                "TUF cache was requested but not used"
-            );
-        }
-    );
-    impl_test!(
-        outdated_local_cache,
-        cache = true,
-        setup = |cache_dir: Option<&Path>| {
-            fs::write(
-                cache_dir.unwrap().join("trusted_root.json"),
-                b"fake trusted root",
-            )
-            .expect("cannot write file to cache dir");
-        },
-        verify = |_, cache_dir: Option<&Path>| {
-            let data = fs::read(cache_dir.unwrap().join("trusted_root.json"))
-                .expect("cannot read trusted_root.json from cache");
+    async fn trust_root(cache: Option<&Path>) -> SigstoreTrustRoot {
+        SigstoreTrustRoot::new(cache)
+            .await
+            .expect("failed to construct SigstoreTrustRoot")
+    }
 
-            assert_ne!(
-                data, b"fake trusted root",
-                "TUF cache was not properly updated"
-            );
-        }
-    );
+    #[rstest]
+    #[tokio::test]
+    async fn trust_root_fetch(#[values(None, Some(cache_dir()))] cache: Option<TempDir>) {
+        let cache = cache.as_ref().map(|t| t.path());
+        let root = trust_root(cache).await;
+
+        verify(&root, cache);
+    }
+
+    #[rstest]
+    #[tokio::test]
+    async fn trust_root_outdated(cache_dir: TempDir) {
+        let trusted_root_path = cache_dir.path().join("trusted_root.json");
+        let outdated_data = b"fake trusted root";
+        fs::write(&trusted_root_path, outdated_data)
+            .expect("failed to write to trusted root cache");
+
+        let cache = Some(cache_dir.path());
+        let root = trust_root(cache).await;
+        verify(&root, cache);
+
+        let data = fs::read(&trusted_root_path).expect("failed to read from trusted root cache");
+        assert_ne!(data, outdated_data, "TUF cache was not properly updated");
+    }
 
     #[test]
     fn test_is_timerange_valid() {


src/verify/verifier.rs
```diff
@@ -0,0 +1,197 @@
...
+            .tbs_certificate
+            .get()
+            .expect("Malformed certificate")
+            .expect("Malformed certificate");
+
+        if !extended_key_usage_ext.0.contains(&ID_KP_CODE_SIGNING) {
+            return Err(VerificationError::CertificateTypeError(
+                "Extended key usage does not contain `code signing`".into(),
+            ));
+        }
```
woodruffw, line 133: This is already covered by `store.verify_cert_with_time` above, so I think it can be removed outright.

src/verify/verifier.rs
```diff
@@ -0,0 +1,197 @@
...
+            .tbs_certificate
+            .get()
+            .expect("Malformed certificate")
+            .expect("Malformed certificate");
+
+        if !key_usage_ext.digital_signature() {
+            return Err(VerificationError::CertificateTypeError(
+                "Key usage is not of type `digital signature`".into(),
+            ));
+        }
```
woodruffw, line 120: This check should be unnecessary, since chain building (above) should assert that the leaf contains a `keyUsage.digitalSignature`.

woodruffw (changes requested):

src/verify/models.rs
```diff
@@ -0,0 +1,222 @@
...
+            let cell = OnceCell::new();
+
+            // TODO(tnytown): Switch to setting if offline when Rekor fetching is implemented.
+            #[allow(clippy::unwrap_used)]
+            cell.set(rekor_entry).unwrap();
+
+            cell
+        } else {
+            Default::default()
+        };
```
woodruffw, line 96: I'm having a little trouble following this: why do we put `LogEntry` into a `OnceCell`, rather than just keeping it as `Option<LogEntry>` with a `None` variant when we're missing an entry?

src/bundle/mod.rs
```diff
@@ -15,21 +15,45 @@
...
+macro_rules! required {
+    ($($base:expr )? ; $first_attr:ident $( . $rest_attrs:ident)* $( , $else_err:expr)?) => {
+        $( $base . )? $first_attr.as_ref()
+            $(
+                .and_then(|v| v.$rest_attrs.as_ref())
+            )*
+        $( .ok_or($else_err) )?
+    }
+}
+pub(crate) use required;
```
woodruffw, line 31: IMO this macro's usages are pretty hard to read -- I think we'd possibly be better off unfolding these inline (or using more bindings to minimize the number of error sites), plus implementing `TryInto` or another conversion trait where appropriate.

src/verify/models.rs
```diff
@@ -0,0 +1,222 @@
...
+            .collect::<Option<Vec<Certificate>>>()?;
+        let [leaf_cert, chain_certs @ ..] = &certs[..] else {
+            return None;
+        };
+
+        if is_leaf(leaf_cert).is_err() {
+            return None;
+        }
+
+        for chain_cert in chain_certs {
```
jasperpatterson, line 140: Should the `chain_certs` be added to the `VerificationMaterials`? It appears they are thrown away right now and chain verification does not currently work if intermediates are present.

src/verify/models.rs
```diff
@@ -0,0 +1,222 @@
...
+        let entry = LogEntry {
+            uuid: "".to_string(),
+            body: log_entry::Body::hashedrekord(canonicalized_body),
+            attestation: None,
+            integrated_time: required!(tlog_entry; integrated_time)?.parse().ok()?,
+            log_i_d: "".into(),
+            log_index: required!(tlog_entry; log_index)?.parse().ok()?,
+            verification: Verification {
+                inclusion_proof: parsed_inclusion_proof,
+                signed_entry_timestamp: required!(; inclusion_promise.signed_entry_timestamp)?
```
jasperpatterson, line 208: I believe this is _not_ required for v0.2+ bundles? https://github.com/sigstore/protobuf-specs/blob/aa56b9a4929364918688fc7eeec3724c9941ddf9/protos/sigstore_rekor.proto#L105

src/crypto/certificate.rs
```diff
@@ -120,6 +120,147 @@ fn verify_expiration(certificate: &Certificate, integrated_time: i64) -> Result<
...
+    //
+    // Any other combination of states is inconsistent and invalid, meaning
+    // that we won't consider the certificate a valid non-CA leaf.
+
+    let ca = match tbs.get::<constraints::BasicConstraints>()? {
+        None => return Ok(false),
+        Some((false, _)) => {
+            // BasicConstraints must be marked as critical, per RFC 5280 4.2.1.9.
+            return Err(SigstoreError::InvalidCertError(
+                "invalid X.509 certificate: non-critical BasicConstraints in CA".to_string(),
```
jasperpatterson, line 206: In `src/verify/models.rs`, you're checking for an `Ok` result from this function, to determine whether what is expected to be an end certificate may unexpectedly be a root:

```rust
if is_root_ca(chain_cert).is_ok() {
    return None;
}
```

However this situation is valid for an end certificate, and not an error:

https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.9
> This extension MAY appear as a critical or non-critical extension in end entity certificates.

jasperpatterson (commented):
I was playing around with these changes as I'm excited to see this functionality merged and noticed a couple little issues in my testing.

src/verify/models.rs
```diff
@@ -0,0 +1,222 @@
...
+        let [leaf_cert, chain_certs @ ..] = &certs[..] else {
+            return None;
+        };
+
+        if is_leaf(leaf_cert).is_err() {
+            return None;
+        }
+
+        for chain_cert in chain_certs {
+            if is_root_ca(chain_cert).is_ok() {
```
jasperpatterson, line 141: I think you need to specifically check the bool here, not just for `Ok`.

src/verify/verifier.rs
```diff
@@ -0,0 +1,174 @@
...
+        else {
+            return Err(VerificationError::CertificateVerificationFailure);
+        };
+
+        debug!("signing certificate chains back to trusted root");
+
+        // TODO(tnytown): verify SCT here, sigstore-rs#326
+
+        // 2) Verify that the signing certificate belongs to the signer.
+        if let Some(err) = policy.verify(&materials.certificate) {
```
tannaurus, line 109: Is there a particular reason this doesn't return `Result<(), PolicyError>`?

src/verify/models.rs
```diff
@@ -0,0 +1,255 @@
...
+    ///
+    /// [VerificationMaterial]: sigstore_protobuf_specs::dev::sigstore::bundle::v1::VerificationMaterial
+    pub fn from_bundle<R: Read>(input: &mut R, bundle: Bundle, offline: bool) -> Option<Self> {
+        let (content, mut tlog_entries) = match bundle.verification_material {
+            Some(m) => (m.content, m.tlog_entries),
+            _ => {
+                error!("bundle missing VerificationMaterial");
+                return None;
+            }
+        };
```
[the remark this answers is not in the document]
tannaurus: From our perspective, returning an error here would be really helpful. We have strict rules on what can and can't be logged, so the errors emitted here won't be captured. Returning error variants here would allow us to decide what gets logged, giving us the visibility that the `error!` had originally intended to provide 😄

src/verify/verifier.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    pub fn verify(
+        &'a self,
+        materials: VerificationMaterials,
+        policy: &impl VerificationPolicy,
+    ) -> VerificationResult {
+        let store = self
+            .cert_pool()
+            .expect("Failed to construct certificate pool");
```
tannaurus, line 65: The outcome of `cert_pool()` is dependent upon the caller's implementation of `Repository`. `Repository` is a public trait that suggests it's ok to return a Result from `fulcio_certs()`

```
pub trait Repository {
    fn fulcio_certs(&self) -> Result<Vec<CertificateDer>>;
    fn rekor_keys(&self) -> Result<Vec<&[u8]>>;
}
```

There is no indication there that returning an error from `fulcio_certs` will cause `verify` to panic. Can this propagate the error instead?

src/verify/models.rs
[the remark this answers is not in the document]
tannaurus: Thanks! We (1Password) have to handle tracing with a lot more caution than many do due the type of data our applications are processing (lots of secrets). For that reason, we make sure only crates we've authored can emit trace events. In addition to that, only data that implements a special "Loggable" trait can be logged, allowing us to tie some compile time guarantees into what is safe and what isn't safe to be logged. 🔒 

This is great for our users (there secrets don't end up in our logs) but it does mean we end up wiring up some extra error handling code when working with third party crates ❤️

src/verify/verifier.rs
```diff
@@ -0,0 +1,174 @@
...
+        // TODO(tnytown): verify SCT here, sigstore-rs#326
+
+        // 2) Verify that the signing certificate belongs to the signer.
+        if let Some(err) = policy.verify(&materials.certificate) {
+            return Err(err)?;
+        }
+        debug!("signing certificate conforms to policy");
+
+        // 3) Verify that the signature was signed by the public key in the signing certificate
+        let Ok(signing_key): SigstoreResult<CosignVerificationKey> =
```
tannaurus, line 115: Please forgive the nits 👀 there's a bunch of places here where `map_err` could be used instead 😄 I don't have much skin in the game of this code base, just thought I'd share because I've found `map_err` to be a nice and reliable friend ❤️

src/verify/verifier.rs
flavio: I agree with @tannaurus

src/verify/verifier.rs
flavio: I agree with @tannaurus

flavio (commented):
That's a big one :)

Sorry for the delay with the review. Overall looks fine to me, I left some comments.

src/verify/verifier.rs
woodruffw: @tannaurus no need to apologize, the nits are greatly appreciated! `map_err` is definitely more idiomatic here.

src/verify/models.rs
```diff
@@ -0,0 +1,255 @@
...
+        if matches!(
+            rekor_entry,
+            TransparencyLogEntry {
+                inclusion_promise: None,
+                inclusion_proof: None,
+                ..
+            }
+        ) {
+            error!("encountered TransparencyLogEntry without any inclusion materials");
+            return None;
```
tannaurus, line 92: This is always going to construct VerificationMaterials, unless it determines it wasn't provided the necessary data to do so. That feels like a case where we should be returning a `Result` instead. What do you think?

src/bundle/mod.rs
```diff
@@ -15,21 +15,110 @@
...
+}
+
+/// Convert log entries returned from Rekor into Sigstore Bundle format entries.
+impl TryFrom<RekorLogEntry> for TransparencyLogEntry {
+    type Error = ();
+
+    fn try_from(value: RekorLogEntry) -> Result<Self, Self::Error> {
+        let canonicalized_body = {
+            let mut body =
+                json_syntax::to_value(value.body).expect("failed to parse constructed Body!");
```
tannaurus, line 94: Can this propagate the error was well? This is implemented on two public structs 😄

src/crypto/certificate.rs
```diff
@@ -120,6 +121,198 @@ fn verify_expiration(certificate: &Certificate, integrated_time: i64) -> Result<
...
+        return Err(CertificateValidationError::VersionUnsupported);
+    }
+
+    // Valid CA certificates must have the following set:
+    //
+    // - `BasicKeyUsage.keyCertSign`
+    // - `BasicConstraints.ca`
+    //
+    // Any other combination of states is inconsistent and invalid, meaning
+    // that we won't consider the certificate a valid non-CA leaf.
```
tannaurus, line 251: This may be my morning brain or perhaps a lack of understanding, but I'm having a hard time understanding the last few words of this sentence. I would expect it to read

> Any other combination of states is inconsistent and invalid, meaning  that we won't consider the certificate a valid CA.

src/crypto/certificate.rs
```diff
@@ -120,62 +121,104 @@ fn verify_expiration(certificate: &Certificate, integrated_time: i64) -> Result<
...
+    #[error("malformed certificate")]
+    Malformed(#[source] x509_cert::der::Error),
+
+    NotLeaf(#[from] NotLeafErrorKind),
+
+    NotCA(#[from] NotCAErrorKind),
+
+    Extension(#[from] ExtensionErrorKind),
+}
+
```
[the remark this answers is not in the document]
tannaurus: For what it's worth, I like it 😄 it opens up the door for strong unit testing.

src/crypto/certificate.rs
```diff
@@ -120,6 +121,198 @@ fn verify_expiration(certificate: &Certificate, integrated_time: i64) -> Result<
...
+    if ca && key_cert_sign {
+        return Ok(());
+    }
+
+    if !(ca || key_cert_sign) {
+        Err(NotCAErrorKind::NotCA)?;
+    }
+
+    // Anything else is an invalid state that should never occur.
+    Err(NotCAErrorKind::Invalid { ca, key_cert_sign })?
```
tannaurus, line 283: In what scenario would this occur? Is there something preventing us from just assert `!ca || !key_cert_sign` on line 278?

```
    if !ca  || !key_cert_sign {
        Err(NotCAErrorKind::NotCA)?;
    }
```

It would be nice to see all validation asserts get checked _before_ we have a path that returns Ok

src/crypto/certificate.rs
```diff
@@ -120,6 +121,198 @@ fn verify_expiration(certificate: &Certificate, integrated_time: i64) -> Result<
...
+    // extensions and have ambiguous CA behavior.
+    if tbs.version != x509_cert::Version::V3 {
+        return Err(CertificateValidationError::VersionUnsupported);
+    }
+
+    // Non-CAs can't possibly be root CAs.
+    is_ca(certificate)?;
+
+    // A certificate that is its own issuer and signer is considered a root CA.
+    if tbs.issuer == tbs.subject {
```
tannaurus, line 309: Tiny suggestion 😄 
```suggestion
    if tbs.issuer != tbs.subject {
      Err(NotCAErrorKind::NotRootCA)?
    }
    
    Ok(())
```

src/crypto/certificate_pool.rs
```diff
@@ -16,34 +16,37 @@
...
+    intermediates: Vec<CertificateDer<'static>>,
 }
 
-impl<'a> CertificatePool<'a> {
+impl CertificatePool {
     /// Builds a `CertificatePool` instance using the provided list of [`Certificate`].
-    pub(crate) fn from_certificates<R, I>(
+    pub(crate) fn from_certificates<'r, 'i, R, I>(
         trusted_roots: R,
         untrusted_intermediates: I,
```
tannaurus, line 35: Something I noticed while digging around: there doesn't appear to be support for `untrusted_intermediates` anywhere. (All invocations of `from_certificates` provide an empty slice). That appears to be the case dating back to November of last year, [here](https://github.com/sigstore/sigstore-rs/pull/305/files)

woodruffw: Yeah, there's no function for untrusted intermediates in the public instance of Sigstore: the trust root (that gets delivered through TUF or however else) constitutes the ultimate trust anchor, and it has the intermediates baked into it directly. 

(The only scenario in which untrusted intermediates make sense is a BYO PKI setup of Sigstore, where the trust root isn't complete and instead each signing cert is accompanied by one or more intermediates for chain construction purposes.)

Cargo.toml
```diff
@@ -109,23 +111,25 @@ rsa = "0.9.2"
 scrypt = "0.11.0"
 serde = { version = "1.0.136", features = ["derive"] }
 serde_json = "1.0.79"
-serde_with = { version = "3.4.0", features = ["base64"] }
```
tannaurus, line 112: This is only used in the `fulcio` module (behind the `fulcio` feature), maybe we can tuck that behind the `fulcio` feature?

src/crypto/verification_key.rs
```diff
@@ -329,6 +329,70 @@ impl CosignVerificationKey {
             }
         }
     }
+
+    /// Verify the signature provided has been actually generated by the given key
+    /// when signing the provided prehashed message.
+    pub fn verify_prehash(&self, signature: Signature, msg: &[u8]) -> Result<()> {
```
tannaurus, line 335: Something we may want to consider as we move forward to 1.0

Many of these functions don't strike me as something that needs to be `pub`. This function is an example of that. The one place this is called constructs this function's signature argument as `Signature::Raw`, but here we're left to account for `Signature::Base64Encoded` despite it seemingly not being a viable code path.

The primary concern I have with exposing functions like these are
1. it makes understanding the crate's usage more difficult than it needs to be, where the crate's documentation is populated with structs and functions that are not useful for any downstream dependents
2. Most critically: it makes everything we do a breaking change. As soon as something is `pub`, we must assume a downstream dependent is using this function with it's current function signature. Abiding by semver semantics becomes very difficult if not impossible.

src/errors.rs
```diff
@@ -109,6 +109,9 @@ pub enum SigstoreError {
     #[error("Certificate pool error: {0}")]
     CertificatePoolError(String),
 
+    #[error("Certificate invalid: {0}")]
+    InvalidCertError(String),
```
tannaurus, line 113: Doesn't look like this is used anywhere 👀

src/fulcio/models.rs
```diff
@@ -31,31 +35,36 @@ where
...
+fn deserialize_inner_detached_sct_signature<'de, D>(de: D) -> Result<Vec<u8>, D::Error>
 where
     D: Deserializer<'de>,
 {
-    let buf = deserialize_base64(de)?;
-
-    serde_json::from_slice(&buf).map_err(serde::de::Error::custom)
+    let buf: Vec<u8> = Base64::<Standard, Padded>::deserialize_as(de)?;
+
+    // The first two bytes indicate the signature and hash algorithms so let's skip those.
```
tannaurus, line 58: Is this not something we should verify is consistent with what we are expecting here? The leading two bytes are dropped from memory here so that source of truth is gone

src/fulcio/models.rs
```diff
@@ -31,31 +35,36 @@ where
...
+    // The next two bytes indicate the size of the signature.
+    let signature_size = u16::from_be_bytes(buf[2..4].try_into().expect("unexpected length"));
```
tannaurus, line 60: could this just be `usize::from_be_bytes`?

src/lib.rs
```diff
@@ -285,8 +285,13 @@ pub mod rekor;
 pub mod tuf;
 
 // Don't export yet -- these types should only be useful internally.
+#[cfg(feature = "bundle")]
```
tannaurus, line 288: Out of curiosity, what is the motivation behind adding this feature flag?

src/verify/policy.rs
```diff
@@ -0,0 +1,298 @@
...
+        let mut extensions = extensions.iter().filter(|ext| ext.extn_id == T::OID);
+
+        // Check for exactly one extension.
+        let (Some(ext), None) = (extensions.next(), extensions.next()) else {
+            return Err(PolicyError::ExtensionNotFound);
+        };
+
+        // Parse raw string without DER encoding.
+        let val = std::str::from_utf8(ext.extn_value.as_bytes())
+            .expect("failed to parse constructed Extension!");
```
tannaurus, line 112: Since these are pub traits/structs, it feels like we should be propagating this error. Is there a reason we're not?

src/verify/verifier.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        // 1) Verify that the signing certificate is signed by the certificate
+        //    chain and that the signing certificate was valid at the time
+        //    of signing.
+        let tbs_certificate = &materials.certificate.tbs_certificate;
+        let issued_at = tbs_certificate.validity.not_before.to_unix_duration();
+        let cert_der: CertificateDer = materials
+            .certificate
+            .to_der()
+            .expect("failed to DER-encode constructed Certificate!")
```
tannaurus, line 97: Is there a particular reason this is safe to do?

src/tuf/mod.rs
```diff
@@ -77,91 +78,129 @@ impl Repository for ManualRepository<'_> {
...
+    /// callers. Those callers should do the following to fetch the trust root ahead of time.
     ///
     /// ```rust
     /// # use tokio::task::spawn_blocking;
     /// # use sigstore::tuf::SigstoreRepository;
     /// # use sigstore::errors::Result;
     /// # #[tokio::main]
     /// # async fn main() -> std::result::Result<(), anyhow::Error> {
-    /// let repo: Result<SigstoreRepository> = spawn_blocking(|| Ok(SigstoreRepository::new(None)?.prefetch()?)).await?;
+    /// let repo: Result<SigstoreRepository> = spawn_blocking(|| Ok(SigstoreRepository::new(None)?)).await?;
```
tannaurus, line 108: nit; now that `prefetch` is removed, this comment is only telling the caller how to use `spawn_blocking` which doesn't feel _super_ helpful

src/tuf/mod.rs
```diff
@@ -77,91 +78,129 @@ impl Repository for ManualRepository<'_> {
...
     /// // Now, get Fulcio and Rekor trust roots with the returned `SigstoreRepository`
     /// # Ok(())
     /// # }
     /// ```
-    pub fn prefetch(self) -> Result<Self> {
-        let _ = self.trusted_root()?;
-        Ok(self)
+    pub fn new(checkout_dir: Option<&Path>) -> Result<Self> {
```
tannaurus, line 113: nit; this is hiding an allocation (line 127). Feels like we can make that clearer to that caller by asking them to provide `Option<PathBuf>` which would allow them to make their own memory management decisions.

src/tuf/mod.rs
```diff
@@ -77,91 +78,129 @@ impl Repository for ManualRepository<'_> {
...
+        };
+
+        let name: TargetName = name.try_into().map_err(Box::new)?;
+        let local_path = checkout_dir.as_ref().map(|d| d.join(name.raw()));
+
+        // Try reading the target from disk cache.
+        let data = if let Some(Ok(local_data)) = local_path.as_ref().map(std::fs::read) {
+            local_data.to_vec()
+        // Try reading the target embedded into the binary.
+        } else if let Some(embedded_data) = constants::static_resource(name.raw()) {
```
tannaurus, line 164: The other invocation of `static_resource` (with `root.json`) panics if it returns `None`. The only place `fetch_target` is invoked is where `name` equals `trusted_root.json`, which from what I can tell should be just as reliably available as as `root.json` is.

What makes `trusted_root.json` different from `root.json`? Is panicing if `root.json` is missing the wrong thing to do?

src/tuf/mod.rs
```diff
@@ -77,91 +78,129 @@ impl Repository for ManualRepository<'_> {
...
+            let mut repo_data = Vec::new();
+            reader.read_to_end(&mut repo_data)?;
+            Ok(repo_data)
+        };
+
+        let name: TargetName = name.try_into().map_err(Box::new)?;
+        let local_path = checkout_dir.as_ref().map(|d| d.join(name.raw()));
+
+        // Try reading the target from disk cache.
+        let data = if let Some(Ok(local_data)) = local_path.as_ref().map(std::fs::read) {
```
tannaurus, line 161: We're asking the caller to tell us where the disk cache is, but we're not communicating our failure to read from that cache. imo this error should be propagated to the caller because it indicates they've provided a bad argument. If we don't tell them that, someone could call `SigstoreRepository::new(whoops_not_a_real_dir)`, thinking it is reading from a local cache but it's actually reading from a remote target.

src/tuf/mod.rs
```diff
@@ -77,91 +78,129 @@ impl Repository for ManualRepository<'_> {
...
+        let data = if Sha256::digest(&data)[..] != target.hashes.sha256[..] {
+            read_remote_target(&name)?
+        } else {
+            data
+        };
+
+        // Write the up-to-date data back to the disk. This doesn't need to succeed, as we can
+        // always fetch the target again later.
+        if let Some(local_path) = local_path {
+            let _ = std::fs::write(local_path, &data);
```
tannaurus, line 191: shouldn't we handle this Result? (re: [my other comment](https://github.com/sigstore/sigstore-rs/pull/311/files#r1510028139))

src/tuf/mod.rs
```diff
@@ -77,91 +78,129 @@ impl Repository for ManualRepository<'_> {
...
+
+        let trusted_root = {
+            let data = Self::fetch_target(&repository, &checkout_dir, "trusted_root.json")?;
+            serde_json::from_slice(&data[..])?
+        };
+
+        Ok(Self { trusted_root })
+    }
+
+    fn fetch_target<N>(
```
tannaurus, line 137: _highly_ personal nit incoming

Lately I've been working with AWS Nitro Enclaves which don't have access to the disk, you're only option for caching is to keep things in memory. While I don't expect that to have a tangible impact on the changes proposed here, I just wanted to call that to attention: someone who wishes to use `SigstoreRepository` would be unable to leverage the support for caching you've worked to implement here. This could be resolved by instead specifying a trait bound of AsyncRead + AsyncWrite.

On the subject of AsyncRead + AsyncWrite, the current cache approach of using std::fs::read/write blocks this thread.  Since we're already using tokio in this crate we could pull in its `fs` feature and use their methods help our downstream dependents a bit.

tannaurus:
I'm afraid that's all I've got for my Saturday morning code review @tnytown . I can circle back around to this review in the coming days 😄 for now I've got myself a new [Playdate](https://play.date/) and a C SDK to go play around with! 🚀

src/verify/verifier.rs
```diff
@@ -0,0 +1,249 @@
...
+        let updater = SigstoreRepository::new(None)?;
+
+        AsyncVerifier::new(Default::default(), updater)
+    }
+}
+
+/// A synchronous Sigstore verifier.
+///
+/// Async callers must use [`AsyncVerifier`]. Async usage of [`Verifier`] will result in a deadlock.
+pub struct Verifier {
```
tannaurus, line 204: At the risk of bikeshedding, I'll just note, with no firm objection nor meaningful skin in this codebase, that I believe having a `sync` module within `verify` is more idiomatic. I personally prefer this because it pulls the context of whether this is sync or async into the "use declaration." Those who are building synchronous projects can find their structs in the `sync` module(s)

I'm interested to hear another person's opinion on this too 🙂

src/verify/verifier.rs
```diff
@@ -0,0 +1,249 @@
...
+                n => hasher.update(&buf[..n]),
+            }
+        }
+
+        self.verify_digest(hasher, bundle, policy, offline).await
+    }
+}
+
+impl AsyncVerifier {
+    /// Constructs an [`AsyncVerifier`] against the public-good trust root.
```
tannaurus, line 193: Is there a thought process behind this not being declared in the other impl? I would expect to see this with some kind of stricter trait bound, but in the absence of that, I'm more curious if there is a motive (even if it's just preferred code style)

src/verify/verifier.rs
```diff
@@ -0,0 +1,249 @@
...
+            .validity
+            .not_before
+            .to_unix_duration()
+            .as_secs();
+        let not_after = tbs_certificate
+            .validity
+            .not_after
+            .to_unix_duration()
+            .as_secs();
+        if !(not_before <= integrated_time && integrated_time <= not_after) {
```
tannaurus, line 155: Is this correct?

I find the named "not_before/after" combined with flipping the bool and `&&` pretty difficult to reason about. 

It feels to me that this could be `not_before > integrated_time && integrated_time > not_after` but there's a non-zero chance my saturday morning not-enough-coffee-brain is just 404ing

Also, it feels like `not_before` should be its own check? "this certificate isn't valid yet" is different than "this certificate was valid for a time but is now expired"

src/verify/policy.rs
```diff
@@ -0,0 +1,298 @@
...
+
+/// The "all of" policy, corresponding to a logical AND between child policies.
+///
+/// An empty list of child policies is considered trivially invalid.
+pub struct AllOf<'a> {
+    children: Vec<&'a dyn VerificationPolicy>,
+}
+
+impl<'a> AllOf<'a> {
+    pub fn new<I: IntoIterator<Item = &'a dyn VerificationPolicy>>(policies: I) -> Option<Self> {
```
tannaurus, line 203: Some general feedback that I hope you're open to

I noticed some inconsistencies in how trait bounds are declared in this PR: sometimes it's `fn my_func<T: SomeTrait>(my_arg: T)` other times it's `fn my_func(my_arg: impl MyTrait)` and other times it's `fn my_func<T>(my_arg: T) -> MyStruct where T: MyTrait`.

I personally like to keep them consistently to the latter (`where` declaration) since that feels like the easiest to read when trait bounds get to be lengthy.

src/verify/verifier.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        let verify_sig =
+            signing_key.verify_prehash(Signature::Raw(&materials.signature), &input_digest);
+        verify_sig.map_err(SignatureErrorKind::VerificationFailed)?;
+
+        debug!("signature corresponds to public key");
+
+        // 4) Verify that the Rekor entry is consistent with the other signing
+        //    materials
+        let Some(log_entry) = materials.tlog_entry(offline, &input_digest) else {
```
tannaurus, line 129: Suggestion 🙂 
`let log_entry = materials.tlog_entry(offline, &input_digest).ok_or(SignatureErrorKind::Transparency)?;`

src/verify/verifier.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// [`VerificationPolicy`].
+    pub async fn verify<R: AsyncRead + Unpin + Send>(
+        &self,
+        mut input: R,
+        bundle: Bundle,
+        policy: &impl VerificationPolicy,
+        offline: bool,
+    ) -> VerificationResult {
+        // arbitrary buffer size, chosen to be a multiple of the digest size.
+        let mut buf = [0u8; 1024];
```
tannaurus, line 174: I understand the motivation here, I like that this makes reading an "input" easier,  but it presumes the input to implement AsyncRead, which may not be true. It also takes a relatively greedy allocation that the downstream application has no control over. 

In my case I've already deserialized a request that contains the payload that my application must re-construct (multiple values appended to each other in a particular order). I _could_ get my payload to implement AsyncRead, but now the intentions of this decision, making the downstream's life easier, has had the unfortunate side effect of making that implementation more strenuous/awkward.   

Would just expecting this to be a `Vec<u8>` make more sense? It solves the problem of a potentially greed allocation and gives the caller the flexibility to work within the constraints of their system / their input's origin.

src/bundle/mod.rs
```diff
@@ -15,21 +15,110 @@
...
+            .transpose()?;
+
+        Ok(TransparencyLogEntry {
+            canonicalized_body,
+            inclusion_promise,
+            inclusion_proof,
+            integrated_time: value.integrated_time,
+            kind_version: Some(KindVersion {
+                kind: "hashedrekord".to_owned(),
+                version: "0.0.1".to_owned(),
```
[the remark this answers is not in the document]
flavio: I'm fine with that, we can tackle that later on

src/verify/verifier.rs
flavio: I like @tannaurus' suggestion

flavio (changes requested):
LGTM, I left some comments.

Aside from my comments, it would be great to see the other comments addressed. For TODO stuff/nodes the outcome of the conversation can be added into the source code as a comment

src/trust/sigstore/mod.rs
```diff
@@ -194,133 +194,170 @@ impl crate::trust::TrustRoot for SigstoreTrustRoot {
...
+            root.rekor_keys().is_ok_and(|v| !v.is_empty()),
+            "no Rekor keys established"
+        );
+        assert!(
+            root.ctfe_keys().is_ok_and(|v| !v.is_empty()),
+            "no CTFE keys established"
+        );
+    }
+
+    macro_rules! impl_test {
```
flavio, line 268: I understand why you're using this macro, but it makes the tests less readable. You can achieve the same result using the [rstest](https://crates.io/crates/rstest) crate, which is already a dev dependency.

Finally, you can use the [`tokio::test`](https://docs.rs/tokio/latest/tokio/attr.test.html) macro to decorate the unit test, which removes more boilerplate from the test.

I can do the changes afterwards, I understand this PR has been ongoing since a long time

flavio (commented):
I left a minor comment, I'm fine with this PR.

flavio:
@tannaurus: can you take one last look at the PR before I merge it?

flavio (approved):
I think we're finally ready to merge it. Sorry for the long time it took!

tannaurus:
I'm so happy to see this merged! Thanks for all the work @tnytown and @jleightcap ! ❤️ 🚀 

Sorry for my delayed response [here](https://github.com/sigstore/sigstore-rs/pull/311#issuecomment-2051111327). I was out of town for team offsite and then came down with a bug when I got back.

PR #312, opened 2023-11-30
Release 0.8.0

lukehinds:
Once merged, I will tag the release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,211 +1,251 @@
+# v0.8.0
+
+## What's Changed
+
+- chore(deps): Update rstest requirement from 0.17.0 to 0.18.1 by @dependabot in https://github.com/sigstore/sigstore-rs/pull/282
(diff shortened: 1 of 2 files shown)


viccuad (changes requested):
`cargo check` fails, it may be that `webpki::extract_trust_anchor` is now `anchor_from_trusted_cert`.

flavio:
There's also an issue with the `rsa` crate being affected by a security issue. We should bump that as well before tagging the next release

flavio:
[This](https://rustsec.org/advisories/RUSTSEC-2023-0071) is the RSA issue, there's no fix yet. Hence the priority should be to fix the broken tests

flavio:
I've rebased to include the latest fixes landed on the `main` branch. I've also updated the version inside of `Cargo.toml` to be `0.8.0` as discussed above

flavio (approved):

viccuad (approved):

PR #365, opened 2024-05-23
fix: allow ManualTrustRoot to have multiple rekor keys

flavio:
`ManualTrustRoot` implements the `TrustRoot` trait, which requires the implemented to have multiple rekor keys. The `ManualTrustRoot` struct has now been updated to handle that, prior to that only one Rekor key was stored inside of a `ManualTrustRoot` instance.

### DIFF
--- a/examples/cosign/verify/main.rs
+++ b/examples/cosign/verify/main.rs
@@ -63,11 +63,11 @@ struct Cli {
 
     /// File containing Rekor's public key (e.g.: ~/.sigstore/root/targets/rekor.pub)
     #[clap(long, required(false))]
-    rekor_pub_key: Option<String>,
+    rekor_pub_keys: Vec<String>,
 
     /// File containing Fulcio's certificate (e.g.: ~/.sigstore/root/targets/fulcio.crt.pem)
     #[clap(long, required(false))]
-    fulcio_cert: Option<String>,
+    fulcio_certs: Vec<String>,
 
     /// The issuer of the OIDC token used by the user to authenticate against Fulcio
     #[clap(long, required(false))]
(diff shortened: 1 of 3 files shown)


src/trust/mod.rs
```diff
@@ -30,7 +30,7 @@ pub trait TrustRoot {
 #[derive(Debug, Default)]
 pub struct ManualTrustRoot<'a> {
     pub fulcio_certs: Option<Vec<CertificateDer<'a>>>,
-    pub rekor_key: Option<Vec<u8>>,
+    pub rekor_keys: Option<Vec<Vec<u8>>>,
```
Xynnn007, line 33: If we have a `Vec` for keys, we could leave out the outer `Option`. When no keys are given, the `Vec` could be empty.

examples/cosign/verify/main.rs
```diff
@@ -236,10 +236,8 @@ async fn fulcio_and_rekor_data(cli: &Cli) -> anyhow::Result<Box<dyn sigstore::tr
 
     let mut data = sigstore::trust::ManualTrustRoot::default();
     if let Some(path) = cli.rekor_pub_key.as_ref() {
-        data.rekor_key = Some(
-            fs::read(path)
-                .map_err(|e| anyhow!("Error reading rekor public key from disk: {}", e))?,
-        );
+        data.rekor_keys = Some(vec![fs::read(path)
```
Xynnn007, line 239: I noticed that the only place to set `rekor_keys` member variable is here. how could we extend the `Vec` here with more keys?

Xynnn007 (commented):
Thanks @flavio for this! There are some naive points unclear to me. Please correct me if I ignored anything

src/trust/mod.rs
flavio: That's a good point. I've changed both `fulcio_certs` and `rekor_keys` to be vectors.

examples/cosign/verify/main.rs
flavio: I've updated the code to reflect the change

flavio:
@Xynnn007 this is ready for another round of review

Xynnn007 (approved):
lgtm. Thanks!

PR #366, opened 2024-05-23
fix docs

flavio:
When building the docs, provide additional information about which feature makes a module/struct/function/method available.

Fix some warnings returned by `cargo doc`. The warnings were caused by broken internal references.

Finally, build the documentation as part of PRs. Fail when something wrong is found while building the docs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,9 @@ license = "Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/sigstore/sigstore-rs"
 
+[package.metadata.docs.rs]
+all-features = true
+
 [features]
 default = ["full-native-tls", "cached-client", "sigstore-trust-root", "bundle"]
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -44,7 +47,13 @@ sign = ["sigstore_protobuf_specs", "fulcio", "rekor", "cert"]
 verify = ["sigstore_protobuf_specs", "fulcio", "rekor", "cert"]
 bundle = ["sign", "verify"]
 
-sigstore-trust-root = ["sigstore_protobuf_specs", "futures-util", "tough", "regex", "tokio/sync"]
+sigstore-trust-root = [
```
Xynnn007, line 50: Not any suggestions to here, just wonder: which tool/how do you re-format the toml format here?

Xynnn007 (approved):
Great! We could foresee this would do help a lot!

Cargo.toml
flavio: I'm using [taplo](https://taplo.tamasfe.dev/), I've got enabled inside of my vim setup

PR #374, opened 2024-06-21
transparency: pull OID constants from `const-oid`

tnytown:
Quick change to resolve a TODO: replace manually defined OIDs with their equivalents in `const-oid`.

CC @flavio


woodruffw (approved):

flavio (approved):
Cool!

PR #381, opened 2024-08-04
fix: Allow empty passwords for encrypted pem files

### LINKED ISSUE #380: Cannot load cosign generated key pairs with empty password
**Description**

I've been working on integrating this crate into my CLI program and I ran into a situation where pem files generated by the cosign tool with an empty password couldn't be loaded. 

```bash
COSIGN_PASSWORD="" COSIGN_YES="true" cosign generate-key-pair
```

I was getting `Failed to parse the key: Unsupported key type`, so I dug into why that was and found that the keys were being processed differently if the password had no length. Since the cosign CLI tool appears to allow the use of an empty password, then I believe that this should change to be in parity.

**Version**
0.9.0

gmpinder:
#### Summary

I've been working on integrating this crate into my CLI program and I ran into a situation where pem files generated by the cosign tool with an empty password couldn't be loaded. I was getting an invalid key type error, so I dug into why that was and found that the keys were being processed differently if the password had no length. Removing that constraint allowed my empty password cosign keys to be loaded properly. I'm hoping this change will be accepted so that this rust implementation is a bit closer to parity with the cosign cli tool.

Closes #380 

#### Release Note

- Allows the use of encrypted PEM files with an empty password to keep in parity with `cosign`
- Allows reading cosign generated encrypted pem files with a different scrypt N parameter

#### Documentation

N/A


gmpinder:
After doing some more testing, there seems to be another spot that's preventing this from working with empty password cosign generated keys. Working on figuring that out

gmpinder:
Ok so I found the other issue, there was apparently a hard check for the `N` param for the kdf encryption of the pem. I've removed it for now since cosign appears to use a value of 65536 which is what was causing the key to not be able to be read.

example test cosign pem:

```json
{"kdf":{"name":"scrypt","params":{"N":65536,"r":8,"p":1},"salt":"/67J9VwY8a6ravOCALfO1P3NGD4Xsk6/M9hNbaXC7+A="},"cipher":{"name":"nacl/secretbox","nonce":"/ht529R6XgnAFmUz/u4jyQTMeom4T6MT"},"ciphertext":"ZJYYk2GQaZgJtHzS0JtUnMhSnQWsnGpD3a5B23weYKodDo2dxUNexEIDp88FRC3jURM4b56EHEcnVUZaDL3cgfkN7Meeo18VYSvZmwI7XhRZs119G6yifi8HqZFbgI3mQwNv0EDp1DzL4wfIZ0pyP+xA/3lNy9myfRd6R3RhGHyIkz5QxxrvZ0uTvGTLNrgKHus/sSn+5ZKl/Q=="}
```

src/crypto/signing_key/kdf.rs
```diff
@@ -122,7 +122,7 @@ impl ScryptKDF {
     /// Check whether the given params is as the default,
     /// to avoid a DoS attack.
     fn check_params(&self) -> Result<()> {
-        match self.params.n == SCRYPT_N && self.params.r == SCRYPT_R && self.params.p == SCRYPT_P {
+        match self.params.r == SCRYPT_R && self.params.p == SCRYPT_P {
```
flavio, line 125: This change isn't related with the main topic of the PR, can you either: move it to a new dedicated PR or move it to a separate commit of this PR?

Moreover, I would prefer to leave the check in place, but update the value we have hard-coded to reflect the new params used by cosign.
I see they are defined [here](https://github.com/sigstore/sigstore-rs/blob/4c7bf8d28be494ab2b76a20c628c6ae63cc3b56b/src/crypto/signing_key/kdf.rs#L35-L42)

While doing that, can you also add a comment pointing to the origin of these values? (a link to the source file and line providing them).

Thanks!

flavio (changes requested):
I'm fine with the changes done to allow empty passwords.

@Xynnn007: would you mind taking a look too?

Xynnn007 (approved):
The logic looks good to me. I relooked inside the cosign code and behavior and ensure that if no password is given, still a encrypted privatekey will be given rather than a plaintext privatekey. The encryption key is derived from `""` with scrypt.

src/crypto/signing_key/kdf.rs
Xynnn007: Btw the current cosign code uses N=65536. See underlying code here https://github.com/secure-systems-lab/go-securesystemslib/blob/main/encrypted/encrypted.go#L206C63-L206C71

loosebazooka:
@woodruffw brought this up in the sigstore clients meeting today. Two points to consider here:
1. Empty password vs unencrypted: Empty password may lead to a assumption of stronger security
2. Should sigstore-rs even be generating long lived keys (and serializing them to PEM)? Should this instead be the job of another tool?

src/crypto/signing_key/kdf.rs
gmpinder: Sure, I can pull that change out into another PR

gmpinder: Another consideration. If I change the value entirely, previously generated PEM files would fail to get loaded. We could try instead having 2 constants and check for the current N value and the N value used by cosign.

gmpinder: Ok so figuring out how to make a new PR from the same fork was too much, so I just split them into 2 separate commits

gmpinder:
> (Thanks @loosebazooka! I forgot to comment on this earlier.)
> 
> > * Empty password vs unencrypted: Empty password may lead to a assumption of stronger security
> 
> To elaborate a bit more on this concern: IMO there's a security risk with enabling people to encrypt with the empty password, insofar as the "encrypted" version is indistinguishable from a version with an actual, strong password. In other words: a user who lacks a full picture of the system may assume that any PEM blob with the `BEGIN ENCRYPTED SIGSTORE PRIVATE KEY` tag is safe to leak/commit to a public source, without realizing that the encryption is effectively nonexistent.
> 
> I think the "fix" involves both `cosign` and `sigstore-rs`:
> 
> 1. It looks like `sigstore-rs` (and maybe `cosign` too?) generates key-bearing PEMs with the `ENCRYPTED` tag unconditionally, even when the private key isn't actually encrypted. When the key is actually unencrypted, the PEM tag probably shouldn't include that marker.
> 2. Rather than encrypting with an empty password (when an empty one is explicitly provided), IMO `sigstore-rs` and `cosign` should refuse the operation and fail.
> 
> (2) is arguably a slippery slope towards "weak" password/entropy testing, which is an anti-pattern. But IMO stopping at empty passwords is perfectly fine; that's the point of having a strong memory-hard hash function like `scrypt` 🙂 

What would happen to the users who have already generated a cosign key with an empty password? Would that just stop working all of a sudden? Would those users then have to create new keys and then update their users to know that they have a different key? I feel like changing that now could break compatibility with previous use cases.

woodruffw:
> What would happen to the users who have already generated a cosign key with an empty password?

There would definitely need to be a transition period here, at least for `cosign` (since AFAIK they follow strict SemVer). Maybe something like this:

1. Loading empty-password keys continues to work, with a warning, until the next major version (3.x), when it becomes a failure.
2. Generating empty-password keys becomes a hard error immediately, since `cosign generate-key-pair` is already allowed to fail as part of its invocation contract

For `sigstore-rs`, I don't think a transition period is needed: the project is still explicitly under active development and is not considered stable. 

Do you think that would accomodate your use case?

gmpinder:
> Do you think that would accomodate your use case?

Not really as we have hundreds of users of our system that has been relying on using those empty cosign passwords. I don't see why breaking user-space to remove something that `cosign` allows in the first place is a good idea. If the `cosign` team wants to make that breaking change, then so be it. Until then, I think `sigstore-rs` should at least accommodate the use-cases that `cosign` has created just so that interoperability can exist, because otherwise we will have to stop using this library altogether so that we don't break our user's builds.

Xynnn007:
For compatibility, we probably can judge the reading logic of the keys by pem header, s.t.

1. if the private key starts with `SIGSTORE ...` and no password is given, will use empty password to derive key to decrypt
2. if the private key starts with `PRIVATE ...` and no password is given, it will be treated as a plaintext key
3. if the private key starts with `SIGSTORE ...` and some password is given, will use the password to derive key to decrypt
4. if the private key starts with `PRIVATE ...` and some password is given, will raise an error to show that the key is not encrypted.

case 2is for compability

gmpinder:
Any update on whether this will be merged in?

src/crypto/signing_key/ecdsa/ec.rs
```diff
@@ -153,6 +153,7 @@ where
                 let ec_seckey = SecretKey::<C>::from_sec1_der(pkcs8.private_key)?;
                 Self::from_private_key(ec_seckey)
             }
+            PRIVATE_KEY_PEM_LABEL if password.is_empty() => Self::from_pem(private_key),
```
Xynnn007, line 156: We'd better add another case for `PRIVATE_KEY_PEM_LABEL if !password.is_empty()` to explicitly show the error. Or it will fall on `Unsupported pem tag {PRIVATE_KEY_PEM_LABEL}`

src/crypto/signing_key/ecdsa/ec.rs
```diff
@@ -405,6 +417,19 @@ mod tests {
         );
     }
 
+    /// This test will try to read an encrypted ecdsa
+    /// private key file, which is generated by `sigstore`.
+    #[test]
+    fn ecdsa_from_encrypted_pem_cosign_empty_password() {
```
Xynnn007, line 423: We might combined this case with `ecdsa_from_encrypted_pem` using `rstest`, s.t.
```rust
#[rstest]
#[case("tests/data/keys/ecdsa_encrypted_private.key", PASSWORD)]
#[case::empty_password("tests/data/keys/cosign_generated_encrypted_empty_private.key", EMPTY_PASSWORD)]
fn ecdsa_from_encrypted_pem(#[case] keypath: &str, #[case] password: &str) {
...
}
```

This would help to keep the tests more dry

src/crypto/signing_key/ecdsa/ec.rs
```diff
@@ -418,6 +443,19 @@ mod tests {
         );
     }
 
+    /// This test will try to encrypt a ecdsa keypair and
+    /// return the pem-encoded contents.
+    #[test]
+    fn ecdsa_to_encrypted_pem_empty_password() {
```
Xynnn007, line 449: ditto with `ecdsa_to_and_from_encrypted_pem`

src/crypto/signing_key/ed25519.rs
```diff
@@ -120,6 +120,7 @@ impl Ed25519Keys {
                 })?;
                 Self::from_key_pair_bytes(key_pair_bytes)
             }
+            PRIVATE_KEY_PEM_LABEL if password.is_empty() => Self::from_pem(encrypted_pem),
```
Xynnn007, line 123: ditto, for `PRIVATE_KEY_PEM_LABEL if !password.is_empty()`

src/crypto/signing_key/ed25519.rs
```diff
@@ -305,6 +304,19 @@ mod tests {
         );
     }
 
+    /// This test will try to read an unencrypted ed25519 with an empty password
+    /// private key file, which is generated by `sigstore`.
+    #[test]
+    fn ed25519_from_unencrypted_pem_empty_password() {
```
Xynnn007, line 310: ditto

src/crypto/signing_key/ed25519.rs
```diff
@@ -330,6 +342,18 @@ mod tests {
         );
     }
 
+    /// This test will try to encrypt a ed25519 keypair and
+    /// return the pem-encoded contents.
+    #[test]
+    fn ed25519_to_encrypted_pem_empty_password() {
```
Xynnn007, line 348: ditto

src/crypto/signing_key/ed25519.rs
```diff
@@ -343,6 +367,32 @@ mod tests {
...
+            .expect("export private key to PEM format failed.");
+        let key = Ed25519Keys::from_encrypted_pem(key.as_bytes(), PASSWORD);
+        assert!(key.is_ok(), "can not create Ed25519Keys from PEM string.");
+    }
+
+    /// This test will generate a Ed25519Keys, encode the private key
+    /// into pem, and decode a new key from the generated pem-encoded
+    /// private key.
+    #[test]
+    fn ed25519_to_and_from_encrypted_pem_empty_password() {
```
Xynnn007, line 387: ditto

src/crypto/signing_key/rsa/keypair.rs
```diff
@@ -101,7 +101,9 @@ impl RSAKeys {
                 })?;
                 Ok(Self::from(private_key))
             }
-
+            RSA_PRIVATE_KEY_PEM_LABEL | PRIVATE_KEY_PEM_LABEL if password.is_empty() => {
+                Self::from_pem(encrypted_pem)
+            }
```
Xynnn007, line 106: Ok, we might also ned to add case for `PRIVATE_KEY_PEM_LABEL if !password.is_empty()` here to tell that not supported `PRIVATE_KEY_PEM_LABEL` with password

src/crypto/signing_key/rsa/keypair.rs
```diff
@@ -298,6 +298,20 @@ mod tests {
         );
     }
 
+    /// This test will try to read an unencrypted rsa with an empty password
+    /// private key file, which is generated by `sigstore`.
+    #[test]
+    fn rsa_from_unencrypted_pem_empty_password() {
```
Xynnn007, line 304: ditto, merge with `rsa_from_unencrypted_pem`

src/crypto/signing_key/rsa/keypair.rs
```diff
@@ -324,6 +338,19 @@ mod tests {
         );
     }
 
+    /// This test will try to encrypt a rsa keypair and
+    /// return the pem-encoded contents. The bit size
+    /// of the rsa key is [`KEY_SIZE`].
+    #[test]
+    fn rsa_to_encrypted_pem_empty_password() {
```
Xynnn007, line 345: ditto, merge with `rsa_to_encrypted_pem`

src/crypto/signing_key/rsa/keypair.rs
```diff
@@ -337,6 +364,32 @@ mod tests {
...
+            .expect("export private key to PEM format failed.");
+        let key = RSAKeys::from_encrypted_pem(key.as_bytes(), PASSWORD);
+        assert!(key.is_ok(), "can not create RSAKeys from PEM string.");
+    }
+
+    /// This test will generate a RSAKeys, encode the private key
+    /// it into pem, and decode a new key from the generated pem-encoded
+    /// private key.
+    #[test]
+    fn rsa_to_and_from_encrypted_pem_empty_password() {
```
Xynnn007, line 384: please merge with `rsa_to_and_from_encrypted_pem`

Xynnn007 (commented):
Thanks for this! More comments about code style.

gmpinder:
Cool, I can work on these this weekend

gmpinder:
Alright, got the tests updated.

src/crypto/signing_key/ecdsa/ec.rs
Xynnn007: This comment seems not to have been fixed. What I mean is to add another case
```
PRIVATE_KEY_PEM_LABEL if !password.is_empty() => Err(SigstoreError::PrivateKeyDecryptError(format!(
                "Unencrypted private key but password provided"
            ))),
```

Or, just raise a warning and ignore the password

src/crypto/signing_key/ed25519.rs
Xynnn007: Also for this, not resolved.

src/crypto/signing_key/ecdsa/ec.rs
gmpinder: My bad, fixing this

gmpinder: Done!

src/crypto/signing_key/ed25519.rs
gmpinder: Fixed

gmpinder:
Got the last few comments resolved

Xynnn007:
@gmpinder Thanks for your adopting the suggestions! Let's see how CI works

Xynnn007:
Good. Passed

Xynnn007 (approved):
LGTM. ptal @flavio

flavio (approved):

gmpinder:
Awesome! Would you be able to get these changes out in a `v0.9.1` patch so my crate is able to make use of it?

flavio:
@gmpinder JFYI, this is going to be part of 0.10.0, which we expect to release soon (see https://github.com/sigstore/sigstore-rs/pull/391)

Issue #384, opened 2024-08-20
Add & update targets in TrustedRoot


fghanmi:
**Description**

The purpose of this issue is to track the changes required to add and update targets in the TrustedRoot.

jku:
I'm leaving this comment here instead of the PR since I think it's not specific to the implementation.

You said in the PR:
> actually, on our side we need to be able to manipulate `trust_root.json` in another Rust tool: `tuftool` (https://github.com/awslabs/tough/tree/develop), since as of now, only `sigstore-rs` manages the new-style targets `trust_root.json`. So, it's the library where this code would fit and would be reusable by other people who have a similar usecase.

I believe up to this point `sigstore-rs` has been 100% a **client** library (I could be wrong, I'm not too familiar at this point).

Modifying the trust root sounds like something only the folks running a sigstore instance would do. I can see how the functionality would be useful for folks running private deployments but... is sigstore-rs is the right place?
* including the code in a client library sounds a little weird -- 600 LOC that that will never be used by the actual client is likely not appealing to maintainers
* From an ecosystem perspective it feels like there should preferably be a single tool to producing/modifying  `trusted_root.json`... It's hard to tell if this PR would move towards that goal or away from it

Maybe this code does make sense in sigstore-rs -- it's true that some tools should exist for this -- but if I was a maintainer, I would like more clarity on the above points

flavio:
I totally agree with @jku. I think this is out of the scope of `sigstore-rs`

jku:
@fghanmi:

have you seen https://github.com/kommendorkapten/trtool -- I think kommendorkapten might be open to improvements if that looks roughly like what you need.

PR #385, opened 2024-08-20
Add functions: add/update/delete Targets in TrustedRoot

fghanmi:
#### Summary

The purpose of this PR is to add the required changes to add and update targets in the TrustedRoot. This will involve implementing two new functions:
`add_target(&mut self, new_target: TargetType, target_name: Target) -> Result<()> `
`update_target(&mut self, target: TargetType, target_name: Target) -> Result<()> `

These new functions improve the flexibility and usability of the TrustedRoot, making it easier to manage targets effectively.

Related issue: https://github.com/sigstore/sigstore-rs/issues/384
#### Release Note

Feature: Add and Update Targets in TrustedRoot

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -155,6 +155,7 @@ hex = "0.4"
 json-syntax = { version = "0.12", features = ["canonicalize", "serde"] }
 tls_codec = { version = "0.4", features = ["derive"] }
 ring = "0.17"
+prost-types = "0.12.6"
 
 [dev-dependencies]
 anyhow = { version = "1.0", features = ["backtrace"] }
--- a/src/trust/sigstore/mod.rs
+++ b/src/trust/sigstore/mod.rs
@@ -44,6 +44,16 @@ pub struct SigstoreTrustRoot {
     trusted_root: TrustedRoot,
 }
 
+pub enum TargetType {
+    Authority(CertificateAuthority),
+    Log(TransparencyLogInstance),
+}
+pub enum Target {
+    CertificateAuthority,
+    TimestampAuthority,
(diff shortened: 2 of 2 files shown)


fghanmi:
Hello @flavio, @danbev, 
Could you please have a look ? 
Thank you in advance!

flavio:
Hi @fghanmi,
sorry for the delay, I was on vacation.

Can you provide some examples about why we would need to offer the ability to alter the contents of the TUF repository?

Currently the code cares only about downloading the official TUF repository or consuming a local checkout, then rely on that to perform different checks.

fghanmi:
> Hi @fghanmi, sorry for the delay, I was on vacation.
> 
> Can you provide some examples about why we would need to offer the ability to alter the contents of the TUF repository?
> 
> Currently the code cares only about downloading the official TUF repository or consuming a local checkout, then rely on that to perform different checks.

Hello @flavio 
actually, on our side we need to be able to manipulate `trust_root.json` in another Rust tool: `tuftool` (https://github.com/awslabs/tough/tree/develop), since as of now, only `sigstore-rs`  manages the new-style targets `trust_root.json`. So, it's the library where this code would fit and would be reusable by other people who have a similar usecase.
In this case, we would be able to use this crate in `tough/tuftool` to manipulate `trust_root.json`.
(a similar change was done on sigstore-go  https://github.com/sigstore/sigstore-go/pull/247)

flavio:
I think changing the contents of the TUF repository is out of the scope of this library. Hence closing.

I can help with getting these changes into the `tough` crate upstream if you want

PR #387, opened 2024-09-06
chore: update cargo audit ignore list

flavio:
A bunch of warnings have popped up. There's nothing we can do about them and they basically no impact on us.

- "RUSTSEC-2024-0370": This is a warning about `proc-macro-errors` being unmaintained. It's a transitive dependency of `sigstore` and `oci-spec`.
- "RUSTSEC-2023-0055": This is a warning about `lexical` having multiple soundness issues. It's a transitive dependency of `sigstore`.

Also, stop ignoring "RUSTSEC-2020-0071" since the issue has been fixed by our transitive dependency.


flavio:
rebased against `main` to include latest fixes

viccuad (approved):

PR #390, opened 2024-09-11
dependency cleanup

flavio:
I noticed our Cargo.toml file needed some love:

- relax dependency versions: a lot of dependencies were pinned in a too strict way. This was especially bad for certain dependencies like `tokio`.
- remove duplicated entry: some dependencies were duplicated between the `dependencies` and `dev-dependencies` sections.
- bump dev-dependencies: some `dev-dependencies` were outdated.
- bump dependencies: some `dependencies` were outdated.

About the `dev-dependencies` and `dependency` bumps. I'm concerned, since dependabot should be taking care of this. I've double checked our configuration and everything seems fine. I'm really puzzled, since we got some dependabot PRs, but we missed a lot of updates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -159,7 +159,6 @@ ring = "0.17"
 [dev-dependencies]
 anyhow = { version = "1.0", features = ["backtrace"] }
 assert-json-diff = "2.0"
-chrono = "0.4"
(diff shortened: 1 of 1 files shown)


viccuad (approved):
nice cleanup :).

PR #392, opened 2024-09-14
Cosign | Add support for client to configure a proxy to pull signatures

Xynnn007:
~~This is still in draft. We will wait for a new release of `oci-client` to include https://github.com/oras-project/rust-oci-client/pull/160~~

#### Summary

In some special cases, client needs to connect a proxy to pull signature via a proxy. This PR mainly accomplishes this.

#### Release Note

* Config changes: Add proxy related items to `ClientConfig`
* Dep changes: update `oci-distribution` to `oci-client`

#### Documentation

TODO

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,13 +67,13 @@ sigstore-trust-root-rustls-tls = [
 ]
 
 cosign-native-tls = [
-  "oci-distribution/native-tls",
+  "oci-client/native-tls",
   "cert",
   "cosign",
   "registry-native-tls",
 ]
 cosign-rustls-tls = [
-  "oci-distribution/rustls-tls",
+  "oci-client/rustls-tls",
   "cert",
   "cosign",
   "registry-rustls-tls",
@@ -81,12 +81,12 @@ cosign-rustls-tls = [
 cosign = ["olpc-cjson"]
 cert = []
 
-registry-native-tls = ["oci-distribution/native-tls", "registry"]
-registry-rustls-tls = ["oci-distribution/rustls-tls", "registry"]
+registry-native-tls = ["oci-client/native-tls", "registry"]
+registry-rustls-tls = ["oci-client/rustls-tls", "registry"]
 registry = ["olpc-cjson"]
 
-mock-client-native-tls = ["oci-distribution/native-tls", "mock-client"]
-mock-client-rustls-tls = ["oci-distribution/rustls-tls", "mock-client"]
+mock-client-native-tls = ["oci-client/native-tls", "mock-client"]
+mock-client-rustls-tls = ["oci-client/rustls-tls", "mock-client"]
 mock-client = []
 
 cached-client = ["cached"]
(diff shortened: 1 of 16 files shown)


maarrat (approved):

maarrat (approved):

maarrat (approved):

maarrat (approved):

maarrat (approved):

Xynnn007:
Hi @flavio Could you please help to take a look upon this PR?

Cargo.toml
```diff
@@ -109,7 +109,7 @@ elliptic-curve = { version = "0.13", features = ["arithmetic", "pem"] }
 futures = "0.3"
 futures-util = { version = "0.3", optional = true }
 lazy_static = "1.5"
-oci-distribution = { version = "0.11", default-features = false, optional = true }
+oci-distribution = { default-features = false, optional = true, package = "oci-client", version = "0.13" }
```
jku, line 112: Is there a reason to leave this rename here? Switching the dependency name to oci-client here (and in the feature definitions) seems like it would be more understandable in the future, but would it break something?

flavio (commented):
Overall LGTM. I left some comments

src/trust/sigstore/mod.rs
```diff
@@ -306,8 +306,8 @@ mod tests {
     fn test_is_timerange_valid() {
         fn range_from(start: i64, end: i64) -> TimeRange {
             let base = chrono::Utc::now();
-            let start: SystemTime = (base + chrono::TimeDelta::seconds(start)).into();
-            let end: SystemTime = (base + chrono::TimeDelta::seconds(end)).into();
+            let start: SystemTime = (base + chrono::TimeDelta::try_seconds(start).unwrap()).into();
+            let end: SystemTime = (base + chrono::TimeDelta::try_seconds(end).unwrap()).into();
```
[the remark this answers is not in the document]
jku: I believe what happened was that chrono did deprecate this, loads of people complained, and they "undeprecated" in the next release. 

So I don't think it's required...  except it is with a specific chrono version.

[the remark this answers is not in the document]
Xynnn007: Thanks folks! I have deleted the commit.

Cargo.toml
Xynnn007: Oh, I originally thought it would bring too many changes. Good suggestion that this would make it not readable. Let me change the rename.

Xynnn007:
Hi @flavio @jku Thanks for the suggestions and I have fixed. PTAL if now it looks good to you

src/registry/mod.rs
```diff
@@ -35,6 +35,8 @@ use crate::errors::Result;
 
 use async_trait::async_trait;
 
+use ::oci_client as oci_client_dep;
```
flavio, line 38: I don't get why you did this rename. Maybe I didn't have enough :coffee: :smile:

flavio (commented):
LGTM, I just left one comment to be clarified before merging

src/registry/mod.rs
Xynnn007: Oh yes, there is a same rust file named `oci_client`, so the `oci_client` would reference not to the crate but the sub module

flavio: Thanks for the explanation

flavio (approved):

PR #403, opened 2024-10-11
cosign: Make verify-blob compatible with sigstore-python

### LINKED ISSUE #288: blob verification fails when used with unsanitized base64 signature strings
**Description**

When verifying a signature passed via a file, trailing newlines are not sanitized.

Using the [`verify_blob`](https://github.com/sigstore/sigstore-rs/blob/main/src/cosign/mod.rs#L163-L174) API with a signature file generated by sigstore-python, [`verify_signature`](https://github.com/sigstore/sigstore-rs/blob/main/src/crypto/verification_key.rs#L263-L267) fails with

```
Error: Invalid byte 10, offset 96.
```

the root issue seems to be differing signature files generated between sigstore-python and cosign (e.g. https://github.com/sigstore/sigstore-rs/tree/main/examples/cosign/verify-blob#sign-the-artifacttxt-file-using-cosign).

(byte 10 == '`\n`')

This is pretty easily fixed with a `.trim()` at the `fs::read_to_string` callsite, but compatibility doesn't seem guaranteed between sigstore clients.

**Version**

https://github.com/sigstore/sigstore-rs/commit/e23046c72ecaf3d4fd63540ed538d8972ac61468

jku:
* This changes cosign::verify_blob() cert argument: it is no longer PEM that is base64 encoded (on top of the encoding that is part of being a PEM), instead it is PEM as already documented.
* verify-blob example now works with certificates that are PEM or base64 PEM
* verify-blob example also trims signature

This makes verify-blob work with python-tuf outputs as well as the cosign outputs.

---

Fixes #288 

Note that _this changes the "cosign API"_: I imagined that this is mostly an internal module but if that is something you want to not change then this approach is probably not good.

### DIFF
--- a/examples/cosign/verify-blob/main.rs
+++ b/examples/cosign/verify-blob/main.rs
@@ -15,6 +15,7 @@
 
 extern crate clap;
 extern crate sigstore;
+use base64::{engine::general_purpose::STANDARD as BASE64_STD_ENGINE, Engine as _};
 use clap::Parser;
 use sigstore::cosign::client::Client;
 use sigstore::cosign::CosignCapabilities;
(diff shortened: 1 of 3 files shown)


flavio (approved):
I think the change is reasonable, we will have to do a minor bump of the release to reflect this change

PR #405, opened 2024-10-11
Lint tests too

jku:
The tests already almost pass clippy so it feels like linting them is a good call:
* Start linting tests
* Allow panic and unwrap in tests
* Fix a few issues from clippy


viccuad (approved):
Approving, thanks!
Will leave the merging to others for now, as I'm afraid this clashes with other bigger PRs.

viccuad:
On second thought, I'm merging this, I would like the new examples to be linted.

Issue #409, opened 2024-10-17
IdentityToken does not expose federated issuer


jku:
**Description**

It would be useful for applications to be able to see the OIDC issuer that was used to get the OIDC token: in the sigstore case this is the issuer that is federated via the sigstore dex instance. This is useful since
1. application may know which identity/issuer will be acceptable in this situation and will be able to cancel before user accidentally signs with incorrect identity 
2. Showing the identity/issuer in the UI may be useful to allow user to verify they are doing the right thing

This federated issuer does not seem to be available in `IdentityToken` currently. Exposing it is slightly more complicate than the identity itself... based on the python implementation it's in `unverified_claims["federated.claims"]["connector_id"]`

https://github.com/sigstore/sigstore-python/blob/main/sigstore/oidc.py#L135

jku:
There's a related issue with IdentityToken: it always uses "email" as the identity of the token -- but some oidc issuers, like GitHub Actions, use "sub" as the identity and Fulcio does respect that...

This leads me to think sigstore-rs does not currently work in GitHub Actions? Have I missed something?

PR #410, opened 2024-10-17
Examples: Add a minimal example of bundle sign/verify

### LINKED ISSUE #406: examples/docs for sigstore::bundle
The `sigstore::bundle` module seems like the "modern high level sigstore API" out of all modules in the project but it is not mentioned in README and has no examples. 

I could try adding an example or two.

jku:
#### Summary

Add example of `sigstore::bundle` use to make it more visible that sigstore-rs does make it easy to use the new bundle format (as well as to automatically update the trust root). More specifically the example uses
 * sigstore::bundle to sign and verify 
 * sigstore::oauth to get a signing token

Fixes #406 

#### Release Note

* examples: Added example of `sigstore::bundle` use

-->

#### Documentation

See included README.md and `--help` output

### DIFF
--- a/examples/bundle/main.rs
+++ b/examples/bundle/main.rs
@@ -1,0 +1,144 @@
+use clap::{Parser, Subcommand};
+use std::fs;
+use std::path::PathBuf;
+use tracing::debug;
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::{fmt, EnvFilter};
+
+use sigstore::bundle::sign::SigningContext;
+use sigstore::bundle::verify::{blocking::Verifier, policy};
+use sigstore::oauth;
+
+#[derive(Parser, Debug)]
+#[clap(about = "Signing and verification example for sigstore::bundle module")]
+struct Cli {
+    /// Enable verbose mode
+    #[arg(short, long)]
+    verbose: bool,
+
+    #[command(subcommand)]
+    command: Commands,
+}
+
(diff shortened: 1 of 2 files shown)


examples/bundle/main.rs
```diff
@@ -0,0 +1,144 @@
...
+            serde_json::to_writer(bundle, &signing_artifact.to_bundle())
+                .expect("Failed to write bundle to file");
+        }
+        Err(e) => {
+            panic!("Failed to sign: {}", e);
+        }
+    }
+    println!(
+        "Created signature bundle {} with identity {}",
+        &bundle_path, email
```
jku, line 97: I'd love to print something like 

```
Signed README.md.sigstore.json. To verify use
    $ cargo run --example bundle verify --identity name@example.com --issuer https://github.com/login/oauth README.md
``` 

But that's not currently possible since the issuer does not seem to be exposed (#409).

examples/bundle/main.rs
```diff
@@ -0,0 +1,144 @@
...
+    );
+}
+
+fn verify(artifact_path: &PathBuf, identity: &str, issuer: &str) {
+    let bundle_path = format!("{}.sigstore.json", artifact_path.display());
```
jku, line 102: Using `display()` here seems incorrect.

viccuad (approved):
Thanks!

flavio (approved):
thanks!

PR #411, opened 2024-10-18
cosign: fix regex dependency import

Xynnn007:
This is a bug fix PR. `regex` dependency is only used in cosign module now. Before this commit, when we only enable `cosign-rustls-tls` feature, the compilation will fail because regex dependency is not enabled.


flavio (approved):
Thanks for the fix

PR #412, opened 2024-10-22
Token fixes

jku:
DRAFT: this is built on top of #410, only two last commits are new.

@tnytown wrt your comments in #409: I've done an initial attempt at improving this. It's draft as mentioned above but I'm also not sure how this API should look like (and am still new with rust)... if you want to take a look at the last commit already and leave suggestions, you're welcome


src/oauth/token.rs
```diff
@@ -28,15 +30,35 @@ pub struct Claims {
...
+            Identity::Email(email) => email.fmt(f),
+        }
+    }
+}
+
 /// A Sigstore token.
 pub struct IdentityToken {
     original_token: String,
     claims: UnverifiedClaims,
+    pub identity: Identity,
```
jku, line 61: The idea here is that in a later PR we'd also add "issuer" field here (which would be the "ultimate issuer": so even if `claims.iss=="https://oauth2.sigstore.dev/auth"`, the issuer field would be the actual identity issuer)

src/bundle/sign.rs
```diff
@@ -85,6 +85,15 @@ impl<'ctx> SigningSession<'ctx> {
         fulcio: &FulcioClient,
         token: &IdentityToken,
     ) -> SigstoreResult<(ecdsa::SigningKey<NistP256>, fulcio::CertificateResponse)> {
+        let identity = match &token.identity {
+            Identity::Sub(_) => {
+                return Err(SigstoreError::IdentityTokenError(
+                    "Non-email identities are not yet supported".to_string(),
+                ))
+            }
```
jku, line 93: interestingly sigstore-python does not seem to handle this case specifically: any identity is slapped in a OID for email...

jku:
> I'm making this a draft again: I made a quick test to try how far the signing on github goes but fulcio does not seem to like the identity token -- I'll figure out if the issue is something in this PR or not before making this ready for review.

Back to ready for review: the IdentityToken is fine, I just messed up GitHubs token-request-token handling in my test.

Speaking of that, I can add add my test for signing-in-a-github-workflow in this PR too, I just figured it does not need to be here. Let me know if you have opinions

src/oauth/token.rs
jku: In a way nothing in IdentityToken is really verified? It's essentially untrusted input (although from a somewhat controlled source, the oidc provider).

Currently `UnverifiedClaims` is a pure deserialization container for the JWT -- I can try adding identity in there somehow but I don't know yet what that will look like.

jku: Yeah I don't hate it...  the annoying part is that `identity()` now returns a Result -- the errors are essentially parsing errors that could have been handled in `try_from()` but now are not so we end up handling errors in more places. 

I will try this to see what it really looks like

jku: Your comment WRT errors sounds good, however I played with this and I don't think IdentityToken construction should succeed if the token does not contain a reasonable identity: it leads to situations where some uses of IdentityToken will succeed and but ones that actually look at the `token.claims.identity()` might fail...

So I think I'll try to address he issue you raised some other way -- maybe naming the field as `token.unverified_identity` or just documenting the token as a whole as semi-untrusted data.

jku:
Latest commit makes it clear that identity claim comes from the OIDC JWT. 

Let me know if you feel strongly about the solution here: I can make identity_claim a method in `Claims` as @tnytown suggested if there are strong opinions -- I didn't do it yet since I felt doing the validation at `try_from()` time is useful.

jku:
FYI in case you'd rather see a more complete picture: https://github.com/jku/sigstore-rs/tree/token-add-issuer-claim contains the second part of this, exposing the "ultimate issue" in IdentityToken.

src/oauth/token.rs
```diff
@@ -104,3 +158,56 @@ impl std::fmt::Display for IdentityToken {
...
+    }
+
+    #[test]
+    fn github_actions_token() {
+        let content = fs::read_to_string("tests/data/tokens/gha-token.txt").unwrap();
+        let identity_token = IdentityToken::try_from(content.as_str()).unwrap();
+        assert_eq!(identity_token.claims.email, None);
+        assert_eq!(
+            identity_token.claims.sub,
+            Some(String::from("repo:sigstore-conformance/extremely-dangerous-public-oidc-beacon:ref:refs/heads/main"))
```
jku, line 197: I'm going to double check if this is correct: It is the token "sub" field...  but it does not seem to be what "bundle verify" use for --identity

jku: I guess that's just how it works? When we ask for the certificate we have no idea what the actual certificate subject is going to be as it's not the identity claim in the OIDC token... I really hope I'm just tired and that's not true

jku: yes, I found the identity handling bits in fulcio and can confirm: **before we get the certificate we ultimately do not know who we are signing as**... exposing the token identity here is still useful since it happens to match the signing identity in the interactive case but it's unfortunately not as useful as I hoped.

PR #415, opened 2024-10-24
Deprecate actions-rs actions

jku:
`actions-rs` actions are unmaintained and archived, we should not be using them. Luckily we weren't really using anything special from them: using `rustup`(that is part of the GitHub image) and `cargo` directly should give us the same results
* Use rustup and cargo instead of specific actions
* Use `Swatinem/rust-cache` to speed up builds

The second part is of course optional. I added it since
* it does speed things up: CI is 1 min faster on wall clock and 4-5 mins faster in job duration
* doing the same thing without a 3rd party action likely leads to annoying debug situations later (rust-cache key calculation seems quite extensive)

### DIFF
--- a/.github/workflows/auto-publish-crates-upon-release.yml
+++ b/.github/workflows/auto-publish-crates-upon-release.yml
@@ -8,10 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       -  uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
-      -  uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1.0.7
-         with:
-           toolchain: stable
-           override: true
+      - name: Rustup
+        run: |
+          rustup install stable
+          rustup override set stable
 
       - name: publish crates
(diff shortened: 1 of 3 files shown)


jku:
If you want to see that the cache works, re-running all CI jobs should work.

flavio:
Instead of using `Swatinem/rust-cache` we could use the GitHub's official `cache` action.

Something like that should work:

```yaml
      - name: Setup Cache
        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
        with:
          path: |
            ~/.cargo/registry
            ~/.cargo/git
            target
          key: ${{ runner.os }}-cargo-clippy-${{ hashFiles('**/Cargo.lock') }}
```

This is taken from [here](https://github.com/containers/oci-spec-rs/blob/main/.github/workflows/ci.yml)

jku:
Sure, I can do that. I wonder if that approach has issues with compiler/toolchain changes though... I suppose if we separate the nightly cache from stable cache these issues might be rare enough.

flavio:
You're right. Let's stick with your original PR. Can you please rebase it against the `main` branch?

jku:
rebased on main, no other changes

viccuad (approved):

flavio (approved):

PR #424, opened 2025-01-23
fix zizmor issues in github actions workflows

bobcallaway:
stops persisting git credentials unnecessarily and drops default permissions to none


flavio (approved):
thanks for the fixes!

PR #431, opened 2025-02-06
Update embedded trust root

### LINKED ISSUE #429: sigstore-rs may be incompatible with root-signing again
See https://github.com/sigstore/root-signing/issues/1431: We (root-signing) may have done the same mistake as happened previously:
* a "keyid" in the tuf repository was not updated  when the key content was modified
* this was not noticed by root-signing test suite because
  * no other sigstore clients require keyids to match the content
  * there is no sigstore-rs test in the test suite
 
I've not confirmed this yet but the report looks correct.

jku:
This is root v12 from Sigstore root-signing (source https://github.com/sigstore/root-signing, published in https://tuf-repo-cdn.sigstore.dev/):
* This is a useful update anyway (since clients now avoid several downloads)
* It is also required since v11 turned out to be incompatible with awslabs/tough

This should fix #429 (once released).


jku:
Of note is that we failed to add sigstore-rs test into root-signing at this time: https://github.com/sigstore/root-signing/issues/1433

I've manually tested and `cargo run --example bundle` now works for both verify and sign

flavio (approved):

PR #437, opened 2025-02-18
oci-client: allow using native cert store with rustls

Xynnn007:
#### Summary

Add support for platform's certs store when using rustls

#### Release Note
Adds a new feature `rustls-tls-native-roots`

#### Documentation

The feature `rustls-tls-native-roots` allows the client to read platform's native certificate store when using rustls suites.

This useful when self-signed registry is used and the certs are loaded in the platform's trusted CA by commands like update-ca-certificates.


viccuad (approved):
LGTM.

Issue #436, opened 2025-02-14
dependency on tough v0.14 pulls in rustls and aws-lc-sys whether you want them or not


dave-tucker:
**Description**

We're using sigstore-rs to perform verification of cosign signatures on the public good instance.
We noticed that in sigstore-rs v0.11 that compilation fails for us on s390x and ppc64le due to a dependency on `aws-lc-sys`.
It turns out that some changes in `tough v0.14` brought in `aws-lc-rs`, `aws-lc-sys` and `rustls` as dependencies, whether you want them or not. 

In our case, we most certainly do not want them as we're using the `native-tls` feature pretty much everywhere and would like to avoid rustls if possible. The fix for this is likely upstream in awslabs/tough, but I wanted to first open an issue as I'm not exactly sure what the correct fix would be.

**Version**

sigstore-rs v0.11

jku:
awslabs/tough#824 I suppose

PR #438, opened 2025-02-18
Tidy up features and dependencies

dave-tucker:
#### Summary

While identifying the root cause of #436 I started looking into the features offered by `sigstore` such that I might avoid the dependency on `tough` entirely - it's marked as optional.

I attempted to use `cargo-hack --every-feature` to test whether the features compiled on their own.
Many features did not. You can reproduce this with:

```
cargo check --no-default-features --features oauth 
cargo check --no-default-features --features fulcio
cargo check --no-default-features --features cosign
```

I then tried `cargo-hack --feature-powerset` to test all combinations of features but it died due to OOM - too many feature combinations.

As such, I took a pass at trying to clean up the feature (and dependency) situation as follows:

1. Dependencies are specified using `default-features=false` and only
       the required features of dependencies are enabled. This avoids
       changes in downstream dependencies default feature-set bringing in
       new dependencies that we don't need.
    
1. I've audited all dependencies to ensure that just those required are
   included in each feature
    
1. All *-native-tls and *-rustls-tls features have been removed and are
   replaced with a single feature for each type of TLS. This correctly
   enables the right features in dependent packages but only if they
   have been selected by another feature (i.e cosign). 
   [This uses the `?` operator added in Rust 1.60](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) but given
   this is over 2 years and 27 stable versions ago I doubt this will be an
   issue.
   
With these changes it was possible to add run `cargo check` in CI for:

- All features individually (with the exception of cosign, full, mock-client and registry which require a `*-tls` feature to compile
- All features combined with `native-tls`
- All features combined with `rustls-tls`

Then to run `cargo test` against `--features full,native-tls` and `--features full,rustls-tls` to ensure that both supported TLS options are properly tested.

#### Release Note

- The following features have been removed: `full-native-tls`,`fulcio-native-tls`,`rekor-native-tls`,`cosign-native-tls`,`mock-client-native-tls`,`sigstore-trust-root-native-tls`,`registry-native-tls`,`full-rustls-tls`,`fulcio-rustls-tls`,`rekor-rustls-tls`,`cosign-rustls-tls`,`mock-client-rustls-tls`,`sigstore-trust-root-rustls-tls`,`registry-rustls-tls`. These features have been replaced with `native-tls` and `rustls-tls` features. To migrate, simply replace `full-native-tls` with `full, native-tls` in your `Cargo.toml`

#### Documentation

No


flavio:
Thanks for this contribution, this is something really interesting that I want to get merged. I'm currently traveling, I'll look closely at that next week

Cargo.toml
```diff
@@ -1,166 +1,238 @@
...
+    "std",
+] }
+regex = { version = "1.10", default-features = false, optional = true }
 reqwest = { version = "0.12", default-features = false, features = [
-  "json",
-  "multipart",
+    "json",
+    "multipart",
+], optional = true }
+ring = { version = "0.17.6", default-features = false, optional = true }
```
viccuad, line 167: we could leave the version unpinned as it was with `0.17`.

viccuad (approved):
LGTM, quite the cleanup, many thanks!
Learned about the `?` operator, looking forward to other reviewers.

.github/workflows/tests.yml
```diff
@@ -4,37 +4,58 @@ name: Continuous integration
...
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           persist-credentials: false
-      - name: Rustup
-        run: |
-          rustup install --profile minimal stable
-          rustup override set stable
+      - uses: dtolnay/rust-toolchain@38b70195107dddab2c7bbd522bcf763bac00963b # stable
       - uses: Swatinem/rust-cache@f0deed1e0edfc6a9be95417288c0e1099b1eeec3 # v2.7.7
+      - uses: taiki-e/install-action@99b53876daf89877c0c7291dfd5033c8c0988612 # cargo-hack
```
flavio, line 15: nitpick, I prefer to be explicit about the version of the `install-action`:

```suggestion
      - uses: taiki-e/install-action@7ea888af71a31437ad4e71c62d021b3bb2728ea9 # v2.49.3
        with:
          tool: cargo-hack
```

I hope I got the indentation right, because the suggest-mode isn't super helpful :sweat_smile:

.github/workflows/tests.yml
```diff
@@ -4,37 +4,58 @@ name: Continuous integration
...
         with:
           persist-credentials: false
-      - name: Rustup
-        run: |
-          rustup install --profile minimal stable
-          rustup override set stable
-          rustup target add wasm32-unknown-unknown
+      - uses: dtolnay/rust-toolchain@38b70195107dddab2c7bbd522bcf763bac00963b # stable
+      - uses: Swatinem/rust-cache@f0deed1e0edfc6a9be95417288c0e1099b1eeec3 # v2.7.7
+      - uses: taiki-e/install-action@99b53876daf89877c0c7291dfd5033c8c0988612 # cargo-hack
```
flavio, line 29: same here

.github/workflows/tests.yml
```diff
@@ -4,37 +4,58 @@ name: Continuous integration
...
+  check-rustls-tls:
+    name: Check rusttls-tls features
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@38b70195107dddab2c7bbd522bcf763bac00963b # stable
+      - uses: Swatinem/rust-cache@f0deed1e0edfc6a9be95417288c0e1099b1eeec3 # v2.7.7
+      - uses: taiki-e/install-action@99b53876daf89877c0c7291dfd5033c8c0988612 # cargo-hack
```
flavio, line 42: same here

Cargo.toml
flavio, line 1: Look like you have been formatting the file using some tool, maybe [taplo](https://taplo.tamasfe.dev/)? If that's the case, can you add a `.taplo.toml` to the PR to ensure the style is automatically enforced?

src/lib.rs
```diff
@@ -234,27 +234,29 @@
...
 //!
-//! - `cached-client`: Enables support for OCI registry client caching.
-//!
-//! - `test-registry`: Enables tests based on a temporary OCI registry.
-//! - `sigstore-trust-root`: Enables support for TUF to request for fulcio certs and rekor public key.
+//! - `cosign`: Enables support for `cosign`.
+//! - `cached-client`: Enables an in-memory cache for the cosign registry client.
+//! - `fulcio`: Enables support for `fulcio`.
+//! - `oauth`: Enables support for `oauth`.
+//! - `registry`: Enables support for `registry`.
```
flavio, line 254: This is a development dependency, I would move it to a separate paragraph and make it explicit

flavio (changes requested):
I like the PR, thanks a lot for the cleanup. I've left some minor requests

flavio:
@dave-tucker on top of my previous comments, could you also take a look at the `Makefile` and propagate some of the changes you did to the GitHub Action over there? It would be great to run at least some sanity checks locally before opening a PR

Cargo.toml
dave-tucker: thanks! done.

.github/workflows/tests.yml
dave-tucker: done.

.github/workflows/tests.yml
dave-tucker: done

Cargo.toml
dave-tucker: yes! pushed a commit which includes taplo config, ci checks for toml formatting + added taplo to `make fmt`

dave-tucker:
@flavio all comments addressed and Makefile updated as requested.

flavio (approved):
Thanks for having done all the requested changes

PR #442, opened 2025-02-25
build(deps): rustls webpki 0.103

fabriziosestito:
#### Summary

This PR resolves the Rustls webpki update issue, which was failing due to a breaking change. Webpki no longer exports `pki-types`, so an explicit `pki-types` dependency was added.  

Reference: https://github.com/rustls/webpki/pull/313

It also adds the `ring` feature, which is now needed because of [this breaking change](https://github.com/rustls/webpki/pull/302).


flavio (approved):
thanks for the fix

PR #467, opened 2025-05-23
feat: add http proxy for ClientConfig

Xynnn007:
This config item allows client to pull signatures from a HTTP registry via a specified proxy.


viccuad (approved):
LGTM, thanks!

PR #476, opened 2025-07-14
update release script to use trusted publishing

bobcallaway:
https://crates.io/docs/trusted-publishing is live, so switching over to it


viccuad (commented):
Thanks!

I believe we must declare:

```yaml
    permissions:
      id-token: write     # Required for OIDC token exchange
```

On the job too, or it will fail on the run (see the linked docs).

flavio (commented):
I'm glad to have this contribution, thanks!

There are some changes needed on the crates.io page, but I don't have access to them. I think we will have to ask for the changes to be done

viccuad (approved):
Approved, thanks!

PR #503, opened 2025-10-16
fix!: handle multiple Rekor keys

### LINKED ISSUE #501: Rekor: Did not find exactly 1 active Rekor key
**Description**

Since a few days ago, we get this error:
```
Did not find exactly 1 active Rekor key
```

Reproducer:
```
use sigstore::trust::{ManualTrustRoot, TrustRoot as _, sigstore::SigstoreTrustRoot};
use std::path::PathBuf;

const RTDIR_TUF: &str = "tuf";

async fn fetch_tuf() {
    print!("Fetching Sigstore TUF data");
    let repo = SigstoreTrustRoot::new(get_tuf_path().as_deref())
        .await
        .unwrap();

    print!("fetching fulcio_certs");
    let fulcio_certs = repo
        .fulcio_certs()
        .expect("Cannot fetch Fulcio certificates from TUF repository")
        .into_iter()
        .map(|c| c.into_owned())
        .collect();

    print!("fulcio certs are: {fulcio_certs:?}");
    print!("repo is: {repo:?}");

    print!("Creating ManualTrustRoot");
    let _ = ManualTrustRoot {
        fulcio_certs,
        rekor_keys: repo
            .rekor_keys()
            .expect("Cannot fetch Rekor keys from TUF repository")
            .iter()
            .map(|k| k.to_vec())
            .collect(),
        ..Default::default()
    };
}

#[tokio::main]
async fn main() {
    fetch_tuf().await;
}

fn get_tuf_path() -> Option<PathBuf> {
    Some(PathBuf::from(RTDIR_TUF))
}
```

```
$ cargo run
(...)
Cannot fetch Rekor keys from TUF repository: TufMetadataError("Did not find exactly 1 active Rekor key")
(...)
```

**Version**

```
$ grep sigstore Cargo.toml 
name = "sigstore-rust"
sigstore = "0.12.1
```

flavio:
Handle multiple Rekor keys found inside of Sigstore's TUF repository.

At bundle verification time, the proper key is found by looking up the key using its ID.

Inside of `trusted_root.json`, each key has a `logId.keyID`, which is a base64 encoded value.
This value must be base64 decoded and then encoded using hex format to build the key ID.

The bundle has `logID` attribute, that is the ID of the key that was used to sign the bundle. This ID must match one of the hex-encoded values built before.

This commit breaks some public APIs of the library, causing us to make a minor release.

The examples have been updated to reflect the new API changes.

The main example about verification (the one bundled inside of our docs) has been changed to build the trust root from the contents of the TUF repository, instead of using local data. This is a more common usage pattern of the library.

Fixes #501

### DIFF
--- a/src/crypto/signing_key/ecdsa/ec.rs
+++ b/src/crypto/signing_key/ecdsa/ec.rs
@@ -78,10 +78,11 @@ use ecdsa::{
     PrimeCurve, SignatureSize, SigningKey,
     hazmat::{DigestPrimitive, SignPrimitive},
 };
+#[allow(deprecated)]
+use elliptic_curve::generic_array::ArrayLength;
 use elliptic_curve::{
     AffinePoint, Curve, CurveArithmetic, FieldBytesSize, PublicKey, Scalar, SecretKey,
     bigint::ArrayEncoding,
-    generic_array::ArrayLength,
     ops::{Invert, Reduce},
     sec1::{FromEncodedPoint, ModulusSize, ToEncodedPoint},
     subtle::CtOption,
@@ -283,6 +284,7 @@ where
 ///
 /// For concrete digest algorithms, please refer to
 /// <https://github.com/RustCrypto/hashes#supported-algorithms>.
+#[allow(deprecated)]
 #[derive(Clone, Debug)]
 pub struct EcdsaSigner<C, D>
 where
@@ -297,6 +299,7 @@ where
     _marker: PhantomData<D>,
 }
 
+#[allow(deprecated)]
 impl<C, D> EcdsaSigner<C, D>
 where
     C: PrimeCurve + CurveArithmetic + AssociatedOid,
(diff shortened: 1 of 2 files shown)


src/trust/sigstore/mod.rs
```diff
@@ -188,24 +205,18 @@ impl crate::trust::TrustRoot for SigstoreTrustRoot {
     /// the local cache if it's not outdated.
     ///
     /// The contents of the local cache are updated when they are outdated.
-    fn rekor_keys(&self) -> Result<Vec<&[u8]>> {
-        let keys: Vec<_> = Self::tlog_keys(&self.trusted_root.tlogs).collect();
+    fn rekor_keys(&self) -> Result<BTreeMap<String, &[u8]>> {
+        let keys: BTreeMap<String, &[u8]> = Self::tlog_keys(&self.trusted_root.tlogs).collect();
 
-        if keys.len() != 1 {
```
flavio: That was the source of the error

viccuad (dismissed):
LGTM! Clean. Thanks for adding the new error and corresponding test case.

PR #511, opened 2025-10-27
Replace json-syntax with serde_json_canonicalizer

### LINKED ISSUE #510: Unmaintained transitive dependencies via json-syntax
**Description**

A dependency of sigstore, the [json-syntax](https://crates.io/crates/json-syntax) crate, seems to be largely unmaintained (no releases in over a year), and its transitive dependencies include `locspan-derive` (unmaintained), `proc-macro-error` (unmaintained), and `syn` v1 (superseded years ago by syn v2, and removing `locspan-derive` will allow sigstore to get rid of syn v1 as a transitive dependency).

It looks like json-syntax is only used in one place, to produce a canonicalized JSON string in `src/bundle/models.rs`. I propose replacing json-syntax with the crate [serde_json_canonicalizer](https://crates.io/crates/serde_json_canonicalizer), an RFC 8785-compatible JSON serializer that's actively maintained and has minimal dependencies aside from serde and serde_json (which are already dependencies of sigstore).

HastD:
#### Summary

The json-syntax crate is only used in a single place and seems to be largely unmaintained, with several unmaintained transitive dependencies. For RFC 8785-compliant JSON canonicalization, we can instead use serde_json_canonicalizer.

Resolves #510.

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/src/rekor/models/search_index_public_key.rs
+++ b/src/rekor/models/search_index_public_key.rs
@@ -31,9 +31,12 @@ impl SearchIndexPublicKey {
 }
 
 ///  The supported pluggable types to sign and upload data
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+#[derive(
+    Default, Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize,
+)]
 pub enum Format {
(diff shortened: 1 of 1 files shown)


viccuad (approved):
LGTM! Many thanks, indeed a good dependency swap.

Pushed a commit to fix an unrelated clippy error for green CI.

PR #513, opened 2025-11-02
feat: Add Merkle tree and Note format foundation for Rekor v2

wolfv:
I have been working on adding bundle v0.3 support to sigstore-rs and have a working version that passes the sigstore-conformance testsuite over at https://github.com/wolfv/sigstore-rs (branch `cms`).
Most of the "work" was done by Claude, using both sigstore-python and sigstore-go as references. 

Getting this merged would be amazing for us and I am happy to spend more time on each PR (also human time, of course)!

This PR adds fundamental data structures needed for Rekor v2 transparency log verification support:

- **src/crypto/merkle.rs**: RFC 6962 Merkle tree operations including leaf hash, node hash, inclusion proof verification, and consistency proof verification. Comprehensive test coverage with test vectors. Tests were partially based on #285 

- **src/crypto/note.rs**: Go note format parsing and verification for Rekor v2 checkpoints. Supports signed notes with multiple signatures, checkpoint verification, and witness signatures.

These modules are pure cryptographic primitives with no dependencies on existing verification code. They serve as the foundation for upcoming RFC 3161 timestamp verification and Bundle v0.3 support.

PR #514, opened 2025-11-02
feat: Add DSSE (Dead Simple Signing Envelope) support

wolfv:
This PR adds support for DSSE envelopes used in in-toto attestations. It's part of our efforts to upstream support for v0.3 bundles (see also #513). The code was mostly written by Claude using sigstore-python and sigstore-go as references.

- **src/bundle/dsse.rs**: DSSE envelope implementation with Pre-Authentication Encoding (PAE) support. Provides DsseEnvelope wrapper around protobuf Envelope with convenient APIs for creating and manipulating DSSE envelopes.

- **src/bundle/intoto.rs**: in-toto Statement support including StatementBuilder, Subject, and predicate handling. Supports creating in-toto attestations with multiple subjects and custom predicates.

- **src/rekor/models/dsse.rs**: Rekor DSSE entry models for submitting DSSE envelopes to the transparency log.

Key Features:
- Correct PAE computation per DSSE spec
- Support for application/vnd.in-toto+json payload type
- Builder pattern for creating statements
- Multiple digest algorithms per subject
- Comprehensive unit tests

This PR lays the groundwork for Bundle v0.3 DSSE verification support and attestation signing capabilities.


baszalmstra (approved):

src/rekor/models/dsse.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub proposed_content: ProposedContent,
+    #[serde(rename = "payloadHash", skip_serializing_if = "Option::is_none")]
+    pub payload_hash: Option<Hash>,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ProposedContent {
+    #[serde(rename = "envelope", skip_serializing_if = "Option::is_none")]
+    pub envelope: Option<String>,
+    #[serde(rename = "verifiers")]
```
Xynnn007: looks like the two `rename` macro target is the same as field name, thus we might not need the rename macros?

src/bundle/intoto.rs
```diff
@@ -0,0 +1,259 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! In-toto attestation statement support for creating DSSE bundles.
+//!
+//! This module provides a builder API for creating in-toto Statement v0.1 attestations
+//! that can be signed and wrapped in DSSE envelopes.
+//!
+//! Note: This implements the v0.1 specification, which is currently used by cosign and Rekor.
+//! See: <https://github.com/in-toto/attestation/blob/main/spec/v0.1.0/statement.md>
```
Xynnn007: This file is not available. It might be the new link `https://github.com/in-toto/attestation/tree/main/spec/v0.1.0#statement`?

src/bundle/intoto.rs
```diff
@@ -0,0 +1,259 @@
...
+        algorithm: impl Into<String>,
+        digest: impl Into<String>,
+    ) -> Self {
+        let mut digest_map = HashMap::new();
+        digest_map.insert(algorithm.into(), digest.into());
+        Self {
+            name: name.into(),
+            digest: digest_map,
+        }
+    }
```
Xynnn007: Just a little thought: 

Could we make this `new()` to only receive `name` parameter? In this way it looks more "natural"
```rust
let subject = Subject::new("name")
    .with_digest(...)
    .with_digest(...);
```

Not sure this makes sense to you

src/bundle/intoto.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Sets the predicate contents for the statement.
+    pub fn predicate(mut self, predicate: serde_json::Value) -> Self {
+        self.predicate = Some(predicate);
+        self
+    }
+
+    /// Builds the statement.
+    ///
+    /// Returns an error if required fields are missing.
+    pub fn build(self) -> Result<Statement, &'static str> {
```
Xynnn007: Could we add a new error enum [here](https://github.com/sigstore/sigstore-rs/blob/main/src/errors.rs) and replace the `Result<Statement, &'static str>` with a unified error/result type?

src/bundle/dsse.rs
```diff
@@ -0,0 +1,281 @@
...
+// limitations under the License.
+
+//! DSSE (Dead Simple Signing Envelope) support for Sigstore bundles.
+//!
+//! This module implements the DSSE specification for creating and verifying
+//! Pre-Authentication Encoding (PAE) used in DSSE signatures.
+//!
+//! See: <https://github.com/secure-systems-lab/dsse/blob/v1.0.0/envelope.md>
+
+use sigstore_protobuf_specs::io::intoto::{Envelope, Signature as DsseSignature};
```
Xynnn007: We might need to re-export the `Envelope` and `DsseSignature` like
```rust
pub use sigstore_protobuf_specs::io::intoto::{Envelope, Signature as DsseSignature};
```

As the function `signatures()` and `as_inner()` returns the two structs, while we cannot expect the caller have imported the dependency `sigstore_protobuf_specs`

Xynnn007 (commented):
Hi @wolfv , nice work! I am also seeking to add support oci v1.1 & bundle v0.3 support in sigstore-rs to work with images signed by cosign v3. Some comments

Xynnn007:
@wolfv Yes, will take a look! 

> if you want to review / give your feedback on those / try them out that would be highly appreciated as well!).

Good! I will take a look

wolfv:
Sorry folks, I don't have the energy to finish this PR here. For our use cases https://github.com/sigstore/sigstore-rust now does a great job.

PR #515, opened 2025-11-02
Pin Rust toolchain to 1.89.0 for consistent builds

wolfv:
Add rust-toolchain.toml to pin the Rust version for consistent clippy/fmt behavior across all contributors and CI runs. 

Stable (1.89+), for example, introduces if-let chain clippy warnings. They are already fixed in the code, meaning it perhaps won't compile anymore with older compilers. I think this should be done deliberately in a certain interval going forward.

We can update Rust version deliberately when we are ready.

## Rationale

Currently CI uses `stable` which floats, causing:
- Clippy warnings that appear/disappear based on developer's Rust version
- Formatting inconsistencies between contributors
- PRs breaking due to new Rust releases unrelated to code changes

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           persist-credentials: false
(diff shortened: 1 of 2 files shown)


viccuad (approved):
Thanks for this contribution!

PR #527, opened 2025-11-17
feat: add merkle tree test suite

wolfv:


viccuad:
Hi, thanks for tackling this.

Sadly, I cannot humanly review a merge commit with 89 changed files nor I can blindly trust it. Hence I elected to perform it on my own plus minimal fixes on https://github.com/sigstore/sigstore-rs/pull/285.

Can you rebase or cherry pick https://github.com/sigstore/sigstore-rs/pull/527/commits/484468e29b8a59dec2308140259452593559cb58 on top of the current state of https://github.com/sigstore/sigstore-rs/pull/285 please?

In addition, that commit will not pass `make test` locally. For that, I would like the merkle tests to be gated under a feature, and a new make target to run them. For now we can keep obtaining the testcase files as a manual step; I would like to keep it simple without submodules, nor importing the history of transparency-dev/merkle, nor copy-pasting them and needing to diff between the repos every time they are updated.

wolfv:
Sorry folks, I won't have the energy to finish this PR. We have added a standalone `sigstore-merkle` crate in https://github.com/sigstore/sigstore-rust which could also be re-used here and passes all the conformance tests.

Issue #307, opened 2023-11-14
Refactor: switch to `ring` library


flavio:
**Description**

A long time ago we moved away from the `ring` crate to a constellation of pure-rust cryptographic libraries. We did the switch because the `ring` library did not build for certain architectures (like s390x and webassembly).

This limitation has been addressed by latest versions of the library. Moreover, we have recently reintroduced the `ring` dependency to implement [TUF trustroots](https://github.com/sigstore/sigstore-rs/pull/305).

Moving back to ring would reduce the list of dependencies we have, making the codebase easier to understand and to maintain.

I think these are the dependencies we should be able to remove:

* ecdsa
* ed25519
* ed25519-dalek
* elliptic-curve
* p256
* p384
* pkcs1
* pkcs8
* rsa

flavio:
@jleightcap: awesome, who should assign this issue to?

viccuad:
> I still think we should drop the pure-rust libraries and support either `ring` or `aws-lc-rs`

It seems that `aws-lc-rs` is [finicky](https://github.com/aws/aws-lc-rs/issues?q=is%3Aissue+is%3Aopen+windows) for Windows builds. Since they are drop-in replacements, I think it would be a good idea to provide a feature for selecting between them.

viccuad:
I'm in for `aws-lc-rs` provided that we can do Windows builds.

PR #534, opened 2025-12-05
refactor(crypto): use awl-lc-rs in ecdsa module.

jvanz:
#### Summary

Updates the code that generates, parses and signs using ECDSA keys to use the aws-lc-rs crate. Therefore, all the references to other crypto libraries has been removed.

The first change the starts to fix the issue #307. 

> [!WARNING]
> Remeber this is **not** a complete change. It's the first of a series of changes in the crypto module to replace crypto libraries in favor of `aws-lc-rs`. See the proposed change [here](https://github.com/sigstore/sigstore-rs/issues/307#issuecomment-3616705700)

#### Release Note

The `EcdsaKeys` struct now requires the callers to algorithm used in the keys they want to parse. This is a required argument to the underling `aws-lc-rs` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ default = ["full", "native-tls"]
 
 bundle = ["sign", "verify"]
 cached-client = ["dep:cached"]
-cert = ["dep:aws-lc-rs", "rustls-webpki/aws-lc-rs"]
+cert = ["rustls-webpki/aws-lc-rs"]
 cosign = [
     "cert",
     "dep:async-trait",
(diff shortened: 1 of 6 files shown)


jvanz:
Just for reference, the `check wasm32-unknown-unknown target` CI is not working due an issue present in the main branch as well. Turns out that `aws-lc-rs` is not building for this target. You can see this by running the following command in the main branch:

```console
cargo check --no-default-features --features wasm,cert --target wasm32-unknown-unknown
```

This was not detected before because the aws-lc-rs in main branch is behind a feature flat which was not being tested. As in this PR the crate is not optional anymore, this start to fail.

flavio:
superseded by https://github.com/sigstore/sigstore-rs/pull/571

PR #571, opened 2026-05-13
feat!: Drop rust crypto & wasm target, align with Sigstore algs

flavio:
Reduce the dependency on Rust Crypto ecosystem. Leverage the aws-lc-rs crate as much as possible.

As a result of that:

- Drop RSA signing variants that use SHA-384 and SHA-512, which are not part of the [Sigstore algorithm registry](https://github.com/sigstore/architecture-docs/blob/main/algorithm-registry.md)
- Drop the WebAssembly target. The aws-lc-rs crate isn't compatible with it. It's usage is unknown and I also wonder how much of that has ever worked.

Some crates of the Rust Crypto ecosystem are still needed, that's mostly because there's no alternative to the x509-cert crate, which brings in some deps from that system.

This PR also bumps some crates of the Rust Crypto ecosystem that are still needed, however we cannot bump all of them because the x509-cert is requiring older versions of them.

**WARNING:** API breakage. CosignVerificationKey, SigningScheme lose variants and gain ECDSA_P521_SHA512_ASN1. enum changes, errors removed


src/crypto/signing_key/ecdsa/ec.rs
```diff
@@ -103,59 +79,93 @@ use crate::{
...
+    /// PKCS#8 DER encoding of the private key (used for cloning / export).
+    pkcs8_der: zeroize::Zeroizing<Vec<u8>>,
+    /// DER-encoded SubjectPublicKeyInfo (cached).
+    spki_der: Vec<u8>,
+}
+
+impl std::fmt::Debug for EcdsaKeys {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EcdsaKeys")
+            .field("curve", &self.curve)
```
Xynnn007, line 133: do we need `spki` also here?

src/crypto/signing_key/ecdsa/ec.rs
```diff
@@ -103,59 +79,93 @@ use crate::{
...
+            .to_vec();
+        Ok(Self {
+            curve,
+            pkcs8_der: zeroize::Zeroizing::new(pkcs8.as_ref().to_vec()),
+            spki_der,
+        })
+    }
+
+    /// Build an `EcdsaKeys` from raw PKCS#8 DER bytes.
+    pub fn from_pkcs8_der(curve: EllipticCurve, der: &[u8]) -> Result<Self> {
```
Xynnn007: Could we judge the curve with only pkcs8 der bytes? i am not sure about this and the aim is to make the API simpler

Xynnn007 (commented):
Thanks @flavio for the replacement!

src/crypto/signing_key/ecdsa/ec.rs
flavio: done with 44e0a52cd5c994a8c76f5cbdc7258e2d59b021db

src/crypto/signing_key/ecdsa/ec.rs
flavio: done with 44e0a52cd5c994a8c76f5cbdc7258e2d59b021db

Xynnn007 (dismissed):
LGTM. Thanks @flavio

Cargo.toml
```diff
@@ -82,7 +85,6 @@ verify = [
     "fulcio",
     "rekor",
 ]
-wasm = ["chrono/wasmbind", "ring/wasm32_unknown_unknown_js"]
```
viccuad, line 85: We need to remove the wasm CI then, as it's failing.

src/crypto/signing_key/rsa/keypair.rs
```diff
@@ -116,157 +139,105 @@ impl RSAKeys {
...
-                }
-                DigestAlgorithm::Sha512 => {
-                    SigStoreSigner::RSA_PKCS1_SHA512(RSASigner::RSA_PKCS1_SHA512(
-                        SigningKey::<sha2::Sha512>::new(private_key),
-                        self.clone(),
-                    ))
-                }
-            },
+        let signer = RSASigner::new(self.clone(), digest_algorithm, padding_scheme);
+        Ok(match (&signer.digest, &signer.padding) {
```
viccuad, line 185: can we be exhaustive on this match for future proofing?

src/crypto/signing_key/ecdsa/ec.rs
```diff
@@ -168,208 +207,123 @@ where
...
     fn sign(&self, msg: &[u8]) -> Result<Vec<u8>> {
-        let mut hasher = D::new();
-        digest::Digest::update(&mut hasher, msg);
-        let (sig, _recovery_id) = self.signing_key.try_sign_digest(hasher)?;
-
-        Ok(sig.to_der().to_bytes().to_vec())
+        let sig = self
+            .inner
+            .sign(&self.rng, msg)
+            .map_err(|e| SigstoreError::PKCS8Error(e.to_string()))?;
```
viccuad: returning this error feels wrong here, as PKCS8 has nothing to do with signing. Same happens in other places, where we return this error.

Given that we are breaking the API, should we bite the bullet and add more granular errors that actually match the case? For example, `SigstoreError::SigningError()`, `SigstoreError::KeyGenerationError()` (instead of the alg errors we had in the past and are being removed in this PR).

viccuad:
LGTM, quite the cleanup, using correct rng, moving from Rust types for keys to using DER. I have some small requests.

Also, given that there's breaking changes in the API of course, worth a shout in the PR description:
- `CosignVerificationKey`, `SigningScheme` lose variants and gain `ECDSA_P521_SHA512_ASN1`.
- enum changes, errors removed

src/crypto/signing_key/rsa/keypair.rs
flavio: we're already covering all the possible matches, since there's only one value for digest (sha256) and two values for the padding scheme (both tested).

I prefer to not ad a "catch all" statement, I prefer the rust compiler fail at compilation time when we will add new values to these enums (if we will ever do that)

Cargo.toml
flavio: fixed with c5220fa646bfb34cc302742003a8c30a1e063904

src/crypto/signing_key/ecdsa/ec.rs
flavio: done with 3d301c8038f9bc27e82f6607a31a053673a9bb89

flavio:
I've pushed one final commit. I've realized we can drop the openssl test dependency with rcgen and, in doing that, make it possible to write a unit test that was blocked by an openssl issue.

src/errors.rs
```diff
@@ -254,6 +254,12 @@ pub enum SigstoreError {
     #[error("Pkcs8 error : {0}")]
     PKCS8Error(String),
```
Xynnn007: I like the new error enum `KeyGenerationError` and `SigningError`, which will tell more user-friendly what happened.
Not sure if the two generic `PKCS8Error` and `PKCS8SpkiError` still works - and if they can be categorized into such types?

Cargo.toml
```diff
@@ -184,13 +184,17 @@ clap = { version = "4.5", default-features = false, features = [
...
+    "aws_lc_rs",
+    "pem",
+] }
 rstest = { version = "0.26", default-features = false }
 serial_test = { version = "3.1", default-features = false }
 tempfile = { version = "3.12", default-features = false }
 testcontainers = { version = "0.27", default-features = false, features = [
     "aws-lc-rs",
 ] }
+time = { version = "0", default-features = false }
```
Xynnn007: The version is unusual. please ensure if this is expected

flavio: `time` is a transitive dependency of rcgen, which doesn't unfortunately exposes it. We specify version "0" to just use the version pulled by rcgen.

I've added a comment explaining that with a3a5835cf1.

Let me know if you're fine with that.

src/errors.rs
flavio: Good point, I removed these two generic errors and introduced more specific ones.

This is done with commit 592ff1ae94, I suggest to review that in "isolation" to reduce "review fatigue"

Cargo.toml
Xynnn007: Thanks for explaination!

Xynnn007 (dismissed):
Thanks for adopting the suggestions! LGTM

src/crypto/verification_key.rs
```diff
@@ -473,20 +469,27 @@ DwIDAQAB
...
+        // supported by CosignVerificationKey (id-Ed448, OID 1.3.101.113). This does
+        // not require generating a real key — we only need the OID to hit the
+        // unsupported-algorithm error path.
+        use const_oid::ObjectIdentifier;
+        use x509_cert::der::asn1::BitString;
+        use x509_cert::spki::AlgorithmIdentifierOwned;
+
+        // id-Ed448: a real algorithm OID, not in the Sigstore registry
+        let id_ed448: ObjectIdentifier = const_oid::db::rfc8410::ID_ED_448;
+        // <review>
```
viccuad: I believe we can drop this comment block.

viccuad (changes requested):
LGTM! Just a small suggestion.

src/crypto/verification_key.rs
flavio: good catch, it's a leftover from me

flavio: fixed and merged into 4f8037ee83314f0ed93088a4d817400233a6ddfd

viccuad (approved):

PR #572, opened 2026-05-18
feat!: cosign v3 signature verification

flavio:
This PR adds support for verifying signatures produced by both SimpleSigning (aka v2 signature) and Sigstore Bundle formats (aka v3 signature).

Sigstore Bundle v0.3 signatures are used by cosign v3's default `--new-bundle-format` mode.

The test suite now has some integration tests that are consuming real world container images signed with v2, v2+v3 and v3 modes. In the future, when this crate will be able to produce also v3 signatures, we can remove the interactions with the remote registries (well, up to a certain degree).

~~Finally, while working on this feature I've uncovered a limitation of the `oci-client` crate. This PR has a workaround that can be removed once https://github.com/oras-project/rust-oci-client/pull/259 is merged upstream (I'm a maintainer over there).~~
**Update:** workaround removed

## Breaking changes

This PR breaks the API used to verify signatures: `CosignCapabilities::trusted_signature_layers` now takes `(auth, source_image)` instead of `(auth, source_image_digest, cosign_image)`.
Triangulation is performed internally. Callers no longer need to call `triangulate()` before verification.


src/cosign/mod.rs
```diff
@@ -84,9 +85,16 @@ pub trait CosignCapabilities {
...
-    /// objects that are associated with the given signature object.
+    /// objects associated with the given image, checking both the cosign
+    /// SimpleSigning (`.sig` tag) format and the Sigstore Bundle (OCI referrers)
+    /// format transparently.
     ///
-    /// Each layer is verified, to ensure it contains legitimate data.
+    /// Triangulation is performed internally: the caller only needs to supply
+    /// the original image reference.  Both signature sources are attempted;
+    /// individual failures are logged as warnings.  Returns an empty vector
+    /// when no layers are found from either source.
```
viccuad, line 95: I would like this to be part of the PR description, with the other breaking changes, in a "breaking changes" paragraph.

(I think our users would benefit greatly from using release-drafter to automatically annotate releases)

src/cosign/client.rs
```diff
@@ -166,6 +185,120 @@ impl Client {
...
+                    &layer.data,
+                    &layer_digest,
+                    source_image_digest,
+                    source_image,
+                    self.fulcio_cert_pool.as_ref(),
+                    self.rekor_pub_keys.as_ref(),
+                ) {
+                    Ok(sl) => layers.push(sl),
+                    Err(e) => {
+                        warn!(error = ?e, "Skipping Sigstore Bundle layer due to error");
```
viccuad, line 293: Question:
Our API depends on returning verified layers and printing warnings for those that we can't obtain nor verify. We cover this with tests in `src/cosign/mod.rs`. I'm scared of refactoring this behavior away. Would it make sense to add unit tests for  `fetch_signature_layers_from_referrers` and `fetch_signature_layers_from_tag`?

src/cosign/client.rs
```diff
@@ -68,30 +69,48 @@ impl CosignCapabilities for Client {
...
     ) -> Result<Vec<SignatureLayer>> {
-        let (manifest, layers) = self.fetch_manifest_and_layers(auth, cosign_image).await?;
-        let image_manifest = match manifest {
-            oci_client::manifest::OciManifest::Image(im) => im,
-            oci_client::manifest::OciManifest::ImageIndex(_) => {
-                return Err(SigstoreError::RegistryPullManifestError {
-                    image: cosign_image.to_string(),
-                    error: "Found a OciImageIndex instead of a OciImageManifest".to_string(),
-                });
+        let (cosign_image, source_image_digest) = self.triangulate(source_image, auth).await?;
```
viccuad, line 73: Question: what happens if we pass a digest directly? Do we fail to triangulate? Do we support both tags, that we can triangulate, and digests?

src/registry/oci_caching_client.rs
```diff
@@ -318,4 +318,26 @@ impl ClientCapabilities for OciCachingClient {
                 error: e.to_string(),
             })
     }
+
+    async fn pull_referrers(
```
viccuad, line 423: This is the same as the non-caching implementation, and maybe we never want to add a `pull_referrers_cached()`.

Thinking on why, I suppose it's because the referrers list is mutable, and new signatures can be added at any time, so we don't want to cache to not have a ~60s window where we may be missing a signature for example. Maybe worth adding a comment on this pull_referrers().

src/cosign/signature_layers.rs
```diff
@@ -180,6 +182,26 @@ impl fmt::Display for SignatureLayer {
...
+/// Compute the DSSE Pre-Authentication Encoding (PAE) bytes.
+///
+/// ```text
+/// PAE(payloadType, payload) =
+///     "DSSEv1" SP LEN(payloadType) SP payloadType SP LEN(payload) SP payload
+/// ```
+///
+/// See <https://github.com/secure-systems-lab/dsse/blob/master/protocol.md>.
+fn compute_pae(payload_type: &str, payload: &[u8]) -> Vec<u8> {
+    let header = format!(
```
viccuad:
```suggestion
    // {} on usize is guaranteed to satisfy the ASCII decimal, no leading zeros, byte LEN()
    let header = format!(
```

src/cosign/intoto.rs
```diff
@@ -0,0 +1,166 @@
...
+    /// Return the SHA-256 digest of the first subject as a hex string (no
+    /// `sha256:` prefix), or an error if the subject list is empty or the
+    /// digest is absent.
+    pub fn subject_sha256_digest(&self) -> Result<String> {
+        self.subject
+            .first()
+            .and_then(|s| s.digest.get("sha256").cloned())
+            .ok_or_else(|| {
+                SigstoreError::UnexpectedError(
+                    "in-toto statement has no subject with a sha256 digest".to_string(),
```
viccuad: It feels to me that these 2 checks (no subject, no sha256) belong to `validate_cosign_v1()`. Or that `validate_cosign_v1()` is underdocumented and could benefit from future proofing. I understand that we do the checks for sigstore things further up the chain, but it's not easy to have in mind the validation against the subset of https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md that sigstore cares about.

For example, we don't care yet about more than `subject[0]`, other checksum algs besides sha256, or `predicate`, but one needs to understand `validate_cosign_v1()` and `SignatureLayer::from_sigstore_bundle()` to see that's ok. I wonder how we can reduce the cognitive load here.

src/bundle/verify/models.rs
```diff
@@ -35,6 +36,28 @@ use x509_cert::{
...
+///
+/// PAE is the canonical byte string that is actually signed in a DSSE envelope.
+/// It encodes both the payload type and payload length to prevent cross-protocol attacks:
+///
+/// ```text
+/// DSSEv1 <len(type)> <type> <len(payload)> <payload>
+/// ```
+///
+/// See <https://github.com/secure-systems-lab/dsse/blob/master/protocol.md>.
+pub(crate) fn compute_pae(payload_type: &str, payload: &[u8]) -> Vec<u8> {
```
viccuad, line 49: This is duplicated with the one in `signature_layers.rs` (plus tests)

src/cosign/signature_layers.rs
```diff
@@ -377,6 +599,327 @@ impl SignatureLayer {
...
+    envelope_json: &[u8],
+    dsse_payload: &[u8],
+    raw_sig: &[u8],
+    cert_der: &[u8],
+) -> Result<()> {
+    use crate::rekor::models::checkpoint::SignedCheckpoint;
+    use crate::rekor::models::inclusion_proof::InclusionProof as RekorInclusionProof;
+    use base64::{Engine as _, engine::general_purpose::STANDARD as base64};
+
+    // 1) Identify the log by its key ID (hex of the raw key_id bytes).
```
viccuad, line 644: there's 3 steps in the function docstring, but 4 in the body. This 1) step is missing in the function docstring.

src/cosign/signature_layers.rs
```diff
@@ -377,6 +599,327 @@ impl SignatureLayer {
...
+        .checkpoint
+        .as_ref()
+        .map(|c| SignedCheckpoint::decode(&c.envelope))
+        .transpose()
+        .map_err(|e| SigstoreError::UnexpectedError(format!("cannot parse checkpoint: {e:?}")))?;
+
+    let proof = RekorInclusionProof::new(
+        proto_proof.log_index,
+        root_hash,
+        proto_proof.tree_size as u64,
```
viccuad: I don't think this cast from i64 to u64 is dangerous, as everything would explode if we would have a negative `tree_size`, or . But maybe it's worth to do a `try_into().map_err(|_|UnexpectedError((format!("tree_size negative"))))` just as the other errors in the function.

src/cosign/signature_layers.rs
```diff
@@ -377,6 +599,327 @@ impl SignatureLayer {
...
+    rekor_key
+        .verify_signature(
+            Signature::Raw(&inclusion_promise.signed_entry_timestamp),
+            &set_payload_canonical,
+        )
+        .map_err(|e| {
+            SigstoreError::UnexpectedError(format!("Sigstore bundle SET verification failed: {e}"))
+        })?;
+
+    // 3) Verify the Merkle inclusion proof (required for v0.3 bundles).
```
viccuad, line 685: This step is quite meaty, I would like to expand the docs.
Something like the following (with the help of an LLM):

> // 3) Verify the Merkle inclusion proof (required for v0.3 bundles).
//
// The proof verification performed by `InclusionProof::verify` chains three
// independent checks; all three are required for the result to be meaningful:
//
//   a. The checkpoint's signature is valid under `rekor_key`; proving that
//      the log itself attested to some (root_hash, tree_size) pair.
//   b. The proof's claimed `(root_hash, tree_size)` matches the *signed*
//      pair from (a); without this cross-check, an attacker could supply a
//      self-consistent Merkle proof against a root the log never signed.
//   c. The Merkle path proves that `canonicalized_body` is the leaf at
//      `log_index` under the now-verified-real root.
//
// Removing any one of (a), (b), (c) breaks the security argument; in
// particular, (b) is the single line `checkpoint.is_valid_for_proof()` in
// `InclusionProof::verify` and is easy to mistake for a redundant assertion.

viccuad: I would also expand the function docs:

> (...)
> /// 2. **Merkle inclusion proof** — offline verification that:
///      * the checkpoint was signed by the log,
///      * the proof's claimed root matches the checkpoint's signed root,
///      * and the entry leaf hash is included in the log tree at the claimed
///            index. All three sub-checks are required: without the second, a
///            self-consistent proof against an attacker-chosen root would verify

src/cosign/signature_layers.rs
```diff
@@ -377,6 +599,327 @@ impl SignatureLayer {
...
+            )
+        })?;
+    let computed_env_hash = hex::encode(Sha256::digest(envelope_json));
+    if computed_env_hash != tlog_env_hash {
+        return Err(SigstoreError::UnexpectedError(format!(
+            "tlog body envelopeHash mismatch: tlog={tlog_env_hash} bundle={computed_env_hash}"
+        )));
+    }
+
+    // --- 3. payload hash ---
```
viccuad: quibble: there's no section 2.
Also, the function docstring lists only 3 checks of the 5.

src/cosign/signature_layers.rs
```diff
@@ -377,6 +599,327 @@ impl SignatureLayer {
...
+        )));
+    }
+
+    let api_version = body
+        .get("apiVersion")
+        .and_then(|v| v.as_str())
+        .ok_or_else(|| {
+            SigstoreError::UnexpectedError("tlog body missing 'apiVersion'".to_string())
+        })?;
+    if api_version != "0.0.1" {
```
viccuad, line 832: I wonder if only checking for the version `0.0.1` is ok with the intoto spec (see [here](https://github.com/in-toto/attestation/blob/main/spec/v1/README.md#parsing-rules)). See "versioning rules":
> Versioning: Each type has a [SemVer2](https://semver.org/) version number and the [TypeURI](https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#TypeURI) reflects the major version number. A message is always semantically correct, but possibly incomplete, when parsed as any other version with the same major version number and thus the same [TypeURI](https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#TypeURI). Minor version changes always follow the monotonic principle. NOTE: 0.X versions are considered major versions.

Supposedly, we should accept `0.0.X` and check the fields ourselves to keep forwards compatiblity if there's a new `0.0.2` intoto served by Rekor. I'm ok as it is right now.. maybe we can add a comment.

src/cosign/signature_layers.rs
```diff
@@ -377,6 +599,327 @@ impl SignatureLayer {
...
+        .and_then(|v| v.as_str())
+        .ok_or_else(|| {
+            SigstoreError::UnexpectedError(
+                "tlog body 'spec.payloadHash.value' missing or not a string".to_string(),
+            )
+        })?;
+    let tlog_payload_alg = spec
+        .pointer("/payloadHash/algorithm")
+        .and_then(|v| v.as_str())
+        .unwrap_or("sha256");
```
viccuad: I dislike this `unwrap_or("sha256")`, can we just emit an error like we do with all the other cases for completeness?

src/cosign/signature_layers.rs
viccuad: I would also like to expand the function docstring. Something like:

> /// Verify that the tlog entry's `canonicalized_body` is a unique fingerprint
/// of the bundle we are presenting.  We verify that:
///
///   1. The entry has kind "dsse" and apiVersion "0.0.1" (anything else has
///      a different body schema and is not supported here).
///   2. `spec.envelopeHash` matches sha256(canonical DSSE envelope JSON).
///   3. `spec.payloadHash` matches sha256(raw DSSE payload bytes).
///   4. `spec.signatures[0].signature` decodes to the same bytes as the bundle's
///      DSSE signature.
///   5. `spec.signatures[0].verifier` is base64(PEM cert) that round-trips to
///      the same DER bytes as the bundle's certificate.
///
/// Without these checks, a valid SET + Merkle proof would only prove that
/// `canonicalized_body` was logged at some point, but they would *not* prove
/// that the logged entry is about *this* bundle.  An attacker holding the
/// bundle's Fulcio cert and any other previously-logged entry by that cert
/// could otherwise present the foreign tlog material as if it were the
/// bundle's signing record, defeating timestamp- and revocation-based
/// policies that depend on the integrated_time being for *this* signature.

src/bundle/verify/models.rs
```diff
@@ -291,4 +402,362 @@ impl CheckedBundle {
...
+            &self.tlog_entry
+        };
+
+        // For DSSE entries the tlog body kind must be "dsse" and the body must be
+        // consistent with the envelope: envelopeHash, payloadHash, signature, and
+        // verifier certificate must all match what we have in the bundle.
+        let actual: serde_json::Value =
+            serde_json::from_slice(&self.tlog_entry.canonicalized_body).ok()?;
+
+        let kind = actual.get("kind").and_then(|v| v.as_str())?;
```
viccuad, line 423: shouldn't we verify here also that apiVersion == 0.0.X as in `verify_bundle_tlog_consistency()`?

src/cosign/intoto.rs
viccuad: We could be even more stringent:  reject `subject.len() != 1`, reject digest objects with more than one algorithm.

src/bundle/verify/models.rs
```diff
@@ -291,4 +402,362 @@ impl CheckedBundle {
...
+
+/// Builds the canonical JSON representation of a DSSE envelope that Rekor uses when
+/// computing `envelopeHash`.  The format matches the JSON serialization produced by
+/// the Rekor `dsse` entry type (v0.0.1): keys are in alphabetical order, binary
+/// fields (`payload`, `sig`) are standard base64-encoded strings.
+///
+/// Returns `None` if the envelope has no signatures (which is invalid for our purposes).
+/// Extract the sha256 digest string from an in-toto Statement v1 payload.
+/// Returns `None` if the payload cannot be parsed or has no sha256 subject.
+fn extract_dsse_subject_sha256(payload: &[u8]) -> Option<String> {
```
viccuad: This function seems duplicated from `InTotoStatementV1::subject_sha256_digest`, could we reuse that one? (with the hardening that I proposed there?)

src/cosign/signature_layers.rs
```diff
@@ -354,6 +376,206 @@ impl SignatureLayer {
...
+        } else {
+            info!("Ignoring certificate in Sigstore bundle: fulcio cert pool not provided");
+            None
+        };
+
+        Ok(SignatureLayer {
+            simple_signing,
+            oci_digest: layer_digest.to_string(),
+            certificate_signature,
+            bundle: None,
```
viccuad: is this `bundle: None` ok?
What happens if the verifier sets `require_rekor_bundle` to `true`? To me it seems that we will fail [in CertificateVerifier::verify()](https://github.com/sigstore/sigstore-rs/blob/83c80ada7f78b65a46f73825ae709a5827f30ad3/src/cosign/verification_constraint/certificate_verifier.rs#L120-L126) even if we shouldn't.

viccuad (changes requested):
I'm ok with the new test-remote-registry being wired into the CI and needing to hit an OCI registry, I see it as needed. It will also catch TUF rotations. In the future we should mirror the images into sigstore org.

I'm still trying to understand the bundle side of the API, FWIW.

src/cosign/client.rs
flavio: fixed with 4bfdd794e1 - it will work just fine, I've added some inline doc explaining that

src/cosign/client.rs
flavio: good point, done with b177817df0

src/bundle/verify/models.rs
flavio: fixed with fe15813e30

src/bundle/verify/models.rs
flavio: Fixed with b0d5aac738

src/bundle/verify/models.rs
flavio: done with cc4a3566ae

src/cosign/signature_layers.rs
flavio: Good feedback, I ended up refactoring how the rekor evidence is handled, see the big refactor inside of 4ea5384968cbfb61e590e07f68ed610fca7824c9

src/cosign/intoto.rs
flavio: done with 35c0c49dad

src/cosign/signature_layers.rs
flavio: fixed with d3a7553304

src/cosign/signature_layers.rs
flavio: fixed with d3a7553304

src/cosign/signature_layers.rs
flavio: fixed with d3a7553304

src/cosign/signature_layers.rs
flavio: good catch, fixed with a50c230803

src/cosign/signature_layers.rs
flavio: fixed with 85a6978d93

src/cosign/signature_layers.rs
flavio: fixed with bcc3fc2cf1

src/registry/oci_caching_client.rs
flavio: actually, I forgot to implement the caching! Done with ed8bfa66f3

src/cosign/signature_layers.rs
flavio: this is now obsolete

src/cosign/mod.rs
flavio: done, I hope that enough

flavio:
@viccuad thanks for the feedback, I've fixed everything

src/registry/oci_caching_client.rs
```diff
@@ -213,6 +213,108 @@ impl PullManifestSettings {
     }
 }
 
+/// Internal struct, used to calculate a unique hash of the pull referrers
```
viccuad, line 244: the new cached block looks good, but it's sitting in the middle of the block of `pull_manifest_cached()`. Could we move `pull_manifest_cached()` before `PullReferrersSettings{}`?

src/cosign/bundle_content.rs
```diff
@@ -0,0 +1,55 @@
...
+///
+/// Cosign supports two provenance formats:
+///
+/// - [`BundleContent::RekorBundle`]: the legacy cosign-specific format,
+///   consisting of a Signed Entry Timestamp (SET) and a payload with body,
+///   integrated time, log index, and log ID.  Stored as OCI annotations on
+///   tag-based cosign signature layers.
+///
+/// - [`BundleContent::SigstoreBundle`]: the protobuf-based Sigstore bundle
+///   format (v0.1–v0.3), carrying a full [`TransparencyLogEntry`](sigstore_protobuf_specs::dev::sigstore::rekor::v1::TransparencyLogEntry).  The entry
```
viccuad: I would like this line to be correctly wrapped.

viccuad (changes requested):
I really like the new `BundleContent`. Still, it seems it should be `pub(crate)` instead of `pub`.

Apart from that, small quibbles but this is ready to go.

flavio:
> I really like the new `BundleContent`. Still, it seems it should be `pub(crate)` instead of `pub`.
> 
> Apart from that, small quibbles but this is ready to go.

It's already `pub(crate)`: https://github.com/flavio/sigstore-rs/blob/5d9230190741591ddbf712b4b982fc2236054c21/src/bundle/verify/models.rs#L169

@viccuad: Am I missing something?

src/cosign/bundle_content.rs
flavio: done with 01cf71a421

src/registry/oci_caching_client.rs
flavio: done with 01cf71a421

viccuad:
> It's already `pub(crate)`: https://github.com/flavio/sigstore-rs/blob/5d9230190741591ddbf712b4b982fc2236054c21/src/bundle/verify/models.rs#L169

Sorry, I thought we could do without making it public, but consumers need it public so they can read `SignatureLayer`, which includes it. Still, we export it twice with 2 paths [here](https://github.com/flavio/sigstore-rs/blob/5d9230190741591ddbf712b4b982fc2236054c21/src/cosign/mod.rs#L53-L54):

`sigstore::cosign::BundleContent` and `sigstore::cosign::bundle_content::BundleContent`. Would be nice to only have the first one to not iincrease our API promises.

flavio:
@viccuad fixed with 31e01d2d1d

viccuad (approved):
LGTM! Thanks for this feature!

PR #580, opened 2026-06-05
crypto: tolerate whitespace in base64 signatures

arpitjain099:
Fixes #550

## Problem

An image signature whose base64 value contains embedded newlines fails verification in sigstore-rs, even though the cosign CLI verifies the same signature successfully.

As reported in #550, a real-world signature pulled from a registry can look like this:

```
signature: Some("MEUCIQDc8DHhEwX/CCESLh\n5RGT+RpA93/Quj9/fi9QpdbZNZPgIgHmhY2+hQLMPQ\npGQbgJr5OqSFOVi3z59+c+wiHNXR2jA=\n")
```

The newlines come from how the signature is stored/transported. cosign and standard base64 tooling tolerate (strip) ASCII whitespace when decoding, so this is a valid signature. sigstore-rs rejected it.

## Root cause

`CosignVerificationKey::verify_signature` and `verify_prehash` in `src/crypto/verification_key.rs` decode the base64 signature with the `STANDARD` base64 engine:

```rust
Signature::Base64Encoded(data) => BASE64_STD_ENGINE.decode(data)?,
```

The `STANDARD` engine treats any byte outside the base64 alphabet (including `\n` and `\r`) as invalid input, so the decode fails and verification returns an error. A valid signature is wrongly rejected.

## Fix

Strip ASCII whitespace from the base64 bytes before decoding, in both decode sites. This matches cosign's behavior. The change is whitespace-only normalization: it does not loosen any other validation. A signature that is malformed for any reason other than embedded whitespace still fails exactly as before (the underlying base64 alphabet/padding/length checks are unchanged), and an invalid signature that happens to decode still fails the cryptographic verify step.

## Before / after

Before the fix, the new test `verify_signature_success_with_embedded_newlines` fails: the signature is identical to the existing `verify_signature_success` fixture but with two `\n` inserted (mid-string and trailing), and verification errors out at the base64 decode step.

After the fix, that test passes, and the existing clean-signature tests (`verify_signature_success`, `verify_rsa_signature`, the failure-path tests) keep passing unchanged.

## Tests

- Added `verify_signature_success_with_embedded_newlines` reusing the existing valid signature/key/message fixture with embedded newlines.
- Existing tests in the module continue to pass, including the negative tests that assert a genuinely malformed signature is still rejected.

## Note on the tradeoff

Whitespace-only normalization is the standard, safe behavior here: it is exactly what cosign and the common base64 decoders do, and it cannot turn an invalid signature into a valid one because the cryptographic verification still runs over the decoded bytes. We deliberately strip only ASCII whitespace and change nothing else about decoding.


src/crypto/verification_key.rs
```diff
@@ -287,6 +310,26 @@ mod tests {
         assert!(outcome.is_ok());
     }
 
+    #[test]
+    fn verify_signature_success_with_embedded_newlines() {
+        // Same valid signature as `verify_signature_success`, but with newlines
+        // embedded inside the base64 string (and a trailing newline). cosign
+        // accepts this; sigstore-rs must too. See sigstore/sigstore-rs#550.
+        let signature = Signature::Base64Encoded(
+            b"MEUCIQD6q/COgzOyW0YH1Dk+CCYSt4uAhm3FDHUwvPI5\n5zwnlwIgE0ZK58ZOWpZw8YVmBapJhBqCfdPekIknimuO0xH8Jh8=\n",
```
viccuad: Could we add the newlines with a constant? That way is more obvious what we are doing, and we don't refactor it away.

src/crypto/verification_key.rs
```diff
@@ -44,6 +44,29 @@ enum EcCurve {
     P521,
 }
 
+/// Decode a base64-encoded signature, tolerating ASCII whitespace.
+///
+/// Some signatures (for example image signatures pulled from a registry) carry
+/// embedded newlines in the base64 value. cosign and the common base64 decoders
```
viccuad: It seems that cosign uses `encoding/json`, which only ignores `\n` and `\r`. Could we amend this to follow that? The rest help catch corruptions.

src/crypto/verification_key.rs
```diff
@@ -44,6 +44,29 @@ enum EcCurve {
...
+/// accept these by ignoring whitespace, so sigstore-rs must too, otherwise a
+/// valid signature is wrongly rejected (see sigstore/sigstore-rs#550).
+///
+/// Only ASCII whitespace is stripped; every other byte is still passed through
+/// to the strict STANDARD engine, so a genuinely malformed signature is rejected
+/// exactly as before.
+fn decode_base64_signature(data: &[u8]) -> Result<Vec<u8>> {
+    if data.iter().any(|b| b.is_ascii_whitespace()) {
```
viccuad: I wouldn't clean every time, as normally there's no need to remove the whitespaces. I would do a `match BASE64_STD_ENGINE.decode(data)`, and on `Err`, clean, and retry the `decode()` again. This leaves the happy path performant and clearer IMHO.

viccuad (changes requested):
Hi, many thanks for tackling this!

In principle this looks ok, I've added some small comments to tighten things.

The blocker for me is that I would like to do this stripping at the source, instead later on. I would prefer to move the new helper function and test into `crypto/mod.rs`  as a new method of `Signature`,  `decode_bytes()`, that does nothing for `Signature::Raw` but performs the decoding for `Signature::Base64Encoded`. 

That way it can be reused by future implementations.

arpitjain099:
Thanks for the detailed review @viccuad! Addressed all three:
- Strips only `\n`/`\r` now, via a `BASE64_IGNORED_NEWLINES` constant, matching cosign's encoding/json behavior - other stray bytes still fail the strict decode so corruption is caught.
- Switched to trying the strict decode first and only stripping + retrying on error, so the unwrapped happy path pays nothing.
- The embedded-newline test now builds its input from explicit newline bytes so the intent is obvious.
All 5 verification_key tests pass locally.

viccuad:
Thanks @arpitjain099, resolved those comments!

Would you like to try tackling the last issue from the previous comment?

> The blocker for me is that I would like to do this stripping at the source, instead later on. I would prefer to move the new helper function and test into crypto/mod.rs as a new method of Signature, decode_bytes(), that does nothing for Signature::Raw but performs the decoding for Signature::Base64Encoded.

arpitjain099:
Done @viccuad - moved the decoding to the source as `Signature::decode_bytes()` in `crypto/mod.rs`: it returns `Raw` bytes unchanged and base64-decodes (with the newline tolerance) for `Base64Encoded`. Both `verify_signature` and `verify_prehash` now just call `signature.decode_bytes()?`, and the decode test moved into `crypto/mod.rs` as a unit test on `decode_bytes`. fmt and tests are green locally.

viccuad (approved):
LGTM! Thanks!

PR #589, opened 2026-06-22
feat:  arbitrary in-toto attestation predicate type support

### LINKED ISSUE #588: Support arbitrary in-toto attestation predicate types in Sigstore Bundle verification
When verifying OCI images that carry Sigstore Bundle v0.3 referrers `sigstore-rs` hard-codes the accepted `predicateType` to `https://sigstore.dev/cosign/sign/v1`. 

This means that images with other valid attestation types — such as SLSA provenance (`https://slsa.dev/provenance/v1`), SBOMs, or any other in-toto predicate — are rejected during `SignatureLayer` construction.

This restriction should be removed so that any in-toto Statement v1 predicate type can be accepted.

## Motivation

Some OCI artifacts are published with attestations that use predicate types other than `cosign/sign/v1`. For example:

```
ghcr.io/webassembly/wasi/http:0.2.12
```

This image carries an SLSA provenance attestation (`https://slsa.dev/provenance/v1`) as a Sigstore Bundle v0.3 OCI referrer. 

Today, `sigstore-rs` rejects this image outright because the `validate_cosign_v1()` method requires `predicateType == "https://sigstore.dev/cosign/sign/v1"`.

This feels unnecessarily restrictive as the in-toto Statement v1 spec is predicate-type-agnostic. 

Removing this restriction enables flexibility around package validation for consumers who need to verify attestation-based trust models (e.g. SLSA provenance, SBOMs, custom attestation types).

## Current Behavior

- `InTotoStatementV1::validate_cosign_v1()` enforces `predicateType == "https://sigstore.dev/cosign/sign/v1"`, causing `SignatureLayer::from_sigstore_bundle()` to reject any bundle with a different predicate type.
- The parsed in-toto statement is validated and then discarded, even if the predicate type check were relaxed, users would have no way to inspect the attestation content after verification.
- There is no `VerificationConstraint` for attestation content, so users cannot express policies like "require SLSA provenance" or "require an attestation with a specific builder ID."

## Proposed Behavior

1. Remove the predicate type restriction
2. Store the attestation on `SignatureLayer`
3. Add an `AttestationVerifier` verification constraint

Attestation content introspection is left entirely to the user — `sigstore-rs` does not interpret or enforce any predicate-specific semantics.

## Use Case

Verifying `ghcr.io/webassembly/wasi/http:0.2.12` which has SLSA provenance but no cosign signature:

```
cargo run --example verify -- \
  --use-sigstore-tuf-data \
  --predicate-type "https://slsa.dev/provenance/v1" \
  ghcr.io/webassembly/wasi/http:0.2.12
```

ryan-surname-p:
#### Summary

Sigstore Bundle v0.3 verification previously hard-coded the accepted in-toto `predicateType` to `https://sigstore.dev/cosign/sign/v1`, rejecting bundles with any other predicate type (e.g. SLSA provenance, SBOMs). 

This means images like `ghcr.io/webassembly/wasi/http:0.2.12` — which carry SLSA provenance attestations cab not be verified using this library.

This PR makes three changes:

1. **Removes the predicate type restriction** — replaces `validate_cosign_v1()` with a general `validate()` method that checks only the structural requirements of an in-toto Statement v1 (correct `_type`, at least one subject with a matching digest). Any valid `predicateType` URI is now accepted.

2. **Stores the parsed attestation on `SignatureLayer`** — the in-toto statement was previously parsed, validated, and discarded. It is now retained on the `SignatureLayer` with accessor methods (`attestation_predicate_type()`, `attestation_predicate()`) so users and verification constraints can inspect the attestation content after verification.

3. **Adds `AttestationVerifier` verification constraint** — a new `VerificationConstraint` implementation that lets users express attestation-based policies: require an attestation is present, optionally match a specific `predicateType`, and optionally supply a closure to validate the `predicate` JSON content. Attestation content introspection is left entirely to the user.

All existing cryptographic verification (SET, Merkle inclusion proof, tlog body consistency, certificate validation) is unchanged — this only affects which predicate types are accepted and what attestation data is exposed after verification.

**To test:** verify an image with a non-cosign attestation type:
```sh
cargo run --example verify -- \
  --use-sigstore-tuf-data \
  --predicate-type "https://slsa.dev/provenance/v1" \
  ghcr.io/webassembly/wasi/http:0.2.12
```

closes #588

#### Release Note

Added support for verifying Sigstore Bundle v0.3 attestations with arbitrary in-toto predicate types.

 Parsed in-toto attestation is now stored on `SignatureLayer` and accessible via `attestation_predicate_type()` and `attestation_predicate()`.

**Compatibility note:** `trusted_signature_layers()` may now return additional `SignatureLayer` objects for bundles with non-cosign predicate types that were previously silently rejected. 

Users who were implicitly relying on only `cosign/sign/v1` layers being returned should add an `AttestationVerifier` constraint with `.with_predicate_type("https://sigstore.dev/cosign/sign/v1")` to restore the previous filtering behavior.

#### Documentation

 Users can reference the `AttestationVerifier` rustdoc and the updated `examples/cosign/verify` example for usage.

### DIFF
--- a/examples/cosign/verify/main.rs
+++ b/examples/cosign/verify/main.rs
@@ -16,8 +16,8 @@
 extern crate sigstore;
 use sigstore::cosign::verification_constraint::cert_subject_email_verifier::StringVerifier;
 use sigstore::cosign::verification_constraint::{
-    AnnotationVerifier, CertSubjectEmailVerifier, CertSubjectUrlVerifier, CertificateVerifier,
-    PublicKeyVerifier, VerificationConstraintVec,
+    AnnotationVerifier, AttestationVerifier, CertSubjectEmailVerifier, CertSubjectUrlVerifier,
+    CertificateVerifier, PublicKeyVerifier, VerificationConstraintVec,
 };
 use sigstore::cosign::{CosignCapabilities, SignatureLayer};
 use sigstore::crypto::SigningScheme;
@@ -99,6 +99,10 @@ struct Cli {
 
     /// Name of the image to verify
     image: OciReference,
+
+    /// Require an in-toto attestation with this predicate type (e.g. https://slsa.dev/provenance/v1)
+    #[clap(long, required(false))]
+    predicate_type: Option<String>,
 
     /// Whether the registry uses HTTP
     #[clap(long)]
@@ -217,6 +221,11 @@ async fn run_app(
         }
     }
 
+    if let Some(predicate_type) = cli.predicate_type.as_ref() {
+        let verifier = AttestationVerifier::new().with_predicate_type(predicate_type);
+        verification_constraints.push(Box::new(verifier));
+    }
+
     let image = &cli.image;
 
     let trusted_layers = client.trusted_signature_layers(auth, image).await?;
@@ -296,6 +305,18 @@ pub async fn main() {
                 match filter_result {
                     Ok(()) => {
                         println!("Image successfully verified");
+                        for layer in &trusted_layers {
+                            if let Some(predicate_type) = layer.attestation_predicate_type() {
+                                println!("  attestation predicate type: {predicate_type}");
+                                if let Some(predicate) = layer.attestation_predicate() {
+                                    println!(
+                                        "  attestation predicate: {}",
+                                        serde_json::to_string_pretty(predicate)
+                                            .unwrap_or_else(|_| format!("{predicate:?}"))
+                                    );
+                                }
+                            }
+                        }
                     }
                     Err(SigstoreVerifyConstraintsError {
                         unsatisfied_constraints,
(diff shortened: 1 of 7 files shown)


wolfv:
You could also try https://github.com/sigstore/sigstore-rust which doesn't have this limitation :)

ryan-surname-p:
> You could also try https://github.com/sigstore/sigstore-rust which doesn't have this limitation :)

Thanks! I was not aware of sigstore-rust but I'll take a look at it today.

wolfv:
Lmk if something is missing for your use case!

ryan-surname-p:
@wolfv, sigstore-rust handles this nicely so I have refactored to use that. Thanks everyone I'll close this as it was just a bandaid to begin with.

PR #610, opened 2026-07-31
crypto: expose keyring module + Keyring type publicly

mlieberman85:
## Summary

`SigningContext::new(fulcio, rekor_config, ctfe_keyring)` at [src/bundle/sign.rs:266](https://github.com/sigstore/sigstore-rs/blob/v0.11.0/src/bundle/sign.rs#L266) accepts a `Keyring` third argument, but the containing `pub(crate) mod keyring` in [src/crypto/mod.rs:181](https://github.com/sigstore/sigstore-rs/blob/v0.11.0/src/crypto/mod.rs#L181) makes the `Keyring` type unreachable from external crates. Callers who want to construct a `SigningContext` without pulling in the `sigstore-trust-root` feature currently have no way to do so — a `pub fn` with a `pub(crate)`-typed parameter.

This PR flips `keyring` to `pub mod` and re-exports `Keyring` at `sigstore::crypto::Keyring` for ergonomic access. Both changes are gated by the existing `cert` feature, matching the module's own gate.

## Motivation

Downstream: [waybill](https://github.com/kusari-sandbox/waybill), an SBOM generator, needs to construct a `SigningContext` with a hand-supplied CTFE keyring for compliance reasons.

`SigningContext::production()` (and `async_production()`) requires the `sigstore-trust-root-*` feature variants, which transitively pull `aws-lc-rs` via [tough](https://github.com/awslabs/tough)'s unconditional dependency on it. waybill's project constitution mandates zero C-native transitives, so the `sigstore-trust-root-*` path is off the table.

With this one-line change (plus a two-line re-export for ergonomics), waybill can:

1. Vendor Sigstore's CTFE public keys as `include_bytes!()` DER SPKI blobs at build time.
2. Construct the CTFE `Keyring` via `Keyring::new([...])`.
3. Compose a `SigningContext` via `SigningContext::new(fulcio, rekor_cfg, ctfe_keyring)`.
4. Sign SBOMs against production Sigstore with zero net-new dependencies and zero C-native code.

Full motivation + audit trail in [waybill's spec 222 §R1](https://github.com/kusari-sandbox/waybill/tree/main/specs/222-sigstore-keyless-signing/research.md) once that branch merges.

## Alternatives considered

- **Fork tough to swap `aws-lc-rs` for `ring`**: much larger surface area, ongoing maintenance burden.
- **Hand-roll the entire signing flow on `FulcioClient` + `rekor::apis` + `sigstore_protobuf_specs` types**: reimplements SCT verification (subtle crypto — Merkle inclusion proofs, log signature validation, timestamp windows). Not the place to reinvent when a vetted API is one visibility flip away.
- **Add a `SigningContext::new_with_vendored_trust_root(...)` helper**: same net effect, but larger API surface than exposing what's already there.

## Compatibility

Additive only. No public API is removed or renamed. The `Keyring` type itself was already `pub struct` — only its containing module's visibility changed. `#[cfg(feature = "cert")]` gate preserved.

## Test plan

- [x] `cargo build --all-features`
- [x] Downstream compilation check: waybill's `sigstore_trust_root.rs` module + api_sanity test that constructs `SigningContext::new(fulcio, rekor_cfg, ctfe_keyring)` compiles + passes under this patch (via `[patch.crates-io]` pin at [waybill Cargo.toml](https://github.com/kusari-sandbox/waybill/blob/222-sigstore-keyless-signing/Cargo.toml))
- [ ] Reviewer: sanity-check that no other `pub fn` accepts a `pub(crate)` type as a parameter (a lint / clippy check would be a good follow-up).

Happy to sign the CLA / adjust the change per maintainer preference (e.g., re-export at `sigstore::bundle::sign::Keyring` instead of `sigstore::crypto::Keyring`).


SantiagoTorres:
>waybill's project constitution mandates zero C-native transitives, so the sigstore-trust-root-* path is off the table.

I see the value in providing this knob, but downside here is that we are effectively allowing for a way to skip a critical portion of the trust path. Would it be a good idea to hide this under a feature flag (e.g., a la `#[cfg(feature='IknowWhatImDoing')]` ) ?

